PR #14, opened 2018-08-24
WIP: Parametric Map Support

NussknackerXXL:
Adding paramtric map support to the viewer.js and enable display2-example to overlay a parametric map over a dicom image series.

Overlay-data is represented by a array-buffer conversion from originally 32bit float to 8bit unsigned int. This conversion is done while the values are rescaled from the previous range to 0 - 255 which makes a colormap integration easy. The example is using a colormap and visual colorbar.

image

public example data: https://drive.google.com/drive/folders/1Yfy0c1dvE7lp4jf99hhZwXDM74Cnc3r4?usp=sharing

### DIFF
--- a/build/dcmjs.js
+++ b/build/dcmjs.js
@@ -50346,7 +50346,7 @@ class Normalizer {
     let toUID = DicomMetaDictionary.sopClassUIDsByName;
     let sopClassUIDMap = {};
     sopClassUIDMap[toUID.CTImage] = CTImageNormalizer;
-	sopClassUIDMap[toUID.ParametricMapStorage] = SEGImageNormalizer;
+    sopClassUIDMap[toUID.ParametricMapStorage] = PMImageNormalizer;
     sopClassUIDMap[toUID.MRImage] = MRImageNormalizer;
     sopClassUIDMap[toUID.EnhancedCTImage] = EnhancedCTImageNormalizer;
     sopClassUIDMap[toUID.LegacyConvertedEnhancedCTImage] = EnhancedCTImageNormalizer;
@@ -50373,6 +50373,7 @@ class Normalizer {
       toUID.EnhancedPETImage,
       toUID.LegacyConvertedEnhancedPETImage,
       toUID.Segmentation,
+      toUID.ParametricMapStorage,
     ];
     return (multiframeSOPClasses.indexOf(sopClassUID) !== -1);
   }
@@ -50757,6 +50758,16 @@ class PETImageNormalizer extends ImageNormalizer {
   }
 }
 
+class PMImageNormalizer extends ImageNormalizer{
+  normalize() {
+    super.normalize();
+    let ds = this.datasets[0]
+    if (ds.BitsAllocated !== 32) {
+      console.error('Only works with 32 bit data, not ' + String(ds.BitsAllocated));
+    }
+  }
+}
+
 class SEGImageNormalizer extends ImageNormalizer {
   normalize() {
     super.normalize();
@@ -50800,6 +50811,7 @@ let normalizers = {
   CTImageNormalizer,
   PETImageNormalizer,
   SEGImageNormalizer,
+  PMImageNormalizer,
   DSRNormalizer
 };
 
--- a/build/dcmjs.js
+++ b/build/dcmjs.js
@@ -1279,6 +1279,15 @@ class OtherByteString extends BinaryRepresentation {
     } */
 }
 
+class OtherFloatString extends BinaryRepresentation{
+     constructor(){
+        super("OF");
+        this.maxLength = null;
+        this.padByte = "00";
+        this.noMultiple = true;
+    }
+}
+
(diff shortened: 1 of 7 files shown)


pieper:
@NussknackerXXL  for reference I am working on another example that uses XMLHttpRequest to download from S3 https://github.com/dcmjs-org/dcmjs/tree/feat/vtk-example  Probably we can clean up this into a file like examples/helpers/s3.js like the dicomweb and tcia code.

NussknackerXXL:
Sure sounds like a good idea. I will try to make your download-routine running within my example, and then extract it in a helper file.

pieper:
Also, just as an FYI we also have a dicomweb-client library work in progress and it occurs to me it might make sense to adapt that so that it can get data from arbitrary URLs like s3 or other static hosting sites, probably also with some kind of json or DICOMDIR support that emulates the ability to do real QIDO.  [Here](https://github.com/dcmjs-org/dicomweb-client/blob/dce62aa5da36c772278706fc35a5fee607c541b9/src/api.js#L48-L122) for example is the request method that could be reused.

This doesn't need to be in your critical path for the PM support @NussknackerXXL , just noting it here.

NussknackerXXL:
Your dicomweb-client code looks also nice and useful. I think i could integrate that as well into my project 👍 

I have now added the download code which works with the image series but if i try to request the ldmap with https://s3.amazonaws.com/IsomicsPublic/SampleData/dicomPM/ldMap_axial_cap.dcm i get a 403. 

@pieper if you would check the link, it could work.

pieper:
Sorry for the delay - I fixed the permission error on the download link.

pieper:
I was thinking rather than copying the promise-ified XMLHttpRequest code everywhere it could be better to use the client library directly like this:

```
<script type="text/javascript" src="https://unpkg.com/dicomweb-client"></script>

      const client = new DICOMwebClient.api.DICOMwebClient({});
      var downloadPromise = client._httpGetApplicationOctetStream( myURL );
```

NussknackerXXL:
Is there something i can do at the moment?

pieper:
Hi @NussknackerXXL  - yes, of course, there are always things to do 😄 Sorry I've probably dropped the ball, so thanks for the reminder.

I haven't looked at the conflicts here but otherwise I don't think it's a problem to merge.  Getting it in the repo will be great.

Once we have the demo merged, the biggest picture thing to do I think is to see if we can integrate this with the DICOM4QI effort (https://dicom4qi.readthedocs.io/en/latest/).  Will the dcmjs code display the sample PM data?

NussknackerXXL:
Update: yes the dcmjs code can display both sample PM data. But i think that here are two use cases to consider. I was looking from an overlay perspective and wrote the example that way.
The sample PM data represents image data itself, which would result in a slightly different example structure. To keep those use cases appart i would create another example which just displays image data like the ADC map. For this use case i would also change the conversion to 16Bit instead of 8Bit used for overlays.

pieper:
Sounds good Tobias - yes, display of PM alone, or with SEG is also important as is the PM  as the overlay.  I hear you might be in Las Palmas, which will give us a chance to meet and discuss with everyone.  Of course it would be great if we can make progress before that too.

pieper:
@NussknackerXXL can we rebase this against the current master to resolve the conflicts?  Would be great to merge this.

NussknackerXXL:
I have made the rebase and after that a few fix and update commits to the changed API, the displayParametricMap example works again. 

It works for the drag and drop way, if you drag the demo data from the s3 storage on the page. Unfortunately, the automatic download from S3 leads to 403 CORS error. I faced a situation like this myself but enabling a CORS browser plugin does not help... But, the files can be downloaded manually..

NussknackerXXL:
Here is another example "displayParametricMap2" which can display standard image data like an ADC from the DICOM4QI effort. It takes some time (like 7 seconds) to convert the 32bit float into 16bit unsigned int image. Automatic download is not covered yet.. 


dcmjs_pm2

NussknackerXXL:
@pieper Any things to do, which prevent this pull request? (Apart from the automatic download)

NussknackerXXL:
Hi Steve!
Thanks, happy new year too.

I found a bug in most of the loaders which prevents loading the last frame of a stack and just displays the frame before the last. Now the scrolling works for me.

The parametric map sample from DICOM4QI could be added to the parametricMap2 example. Maybe there is a not zipped version?

pieper:
Thanks for adding addressing this 👍 

The example is live now https://dcmjs-org.github.io/dcmjs/examples/displayParametricMap/index.html

Let's create newer smaller topics for additional tweaks.

pieper:
I did a few more edits on master and published this all to npm.

Both demos are live now.
https://dcmjs-org.github.io/dcmjs/examples/displayParametricMap2/index.html

Also I found the performance issue and it was actually in the colormap creation.  I switched from insertColor to setColor and everything is very zippy now.

PR #17, opened 2018-10-12
WIP: Add Adapters / Utilities for Imaging Measurement Structured Report input / output

swederik:


fedorov:
Great to see progress in this direction! I would like to review how you guys are approaching this implementation. When were you planning to merge?

Also, is there some kind of a "quick start" guideline how to set things up to be able to try this out locally?

swederik:
It's still in development but I'd be happy to walk you through on a call. I think we plan to merge it once we're happy with the length measurement, and then get back to the other tools later.

I'll see if we can add Netlify for a live link so you don't have to clone and build it. If you want to run it locally, you can just run ```npm run build``` and then check the 'linearMeasurements' example.

fedorov:
http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7470.html

src/utilities/TID300/index.js
```diff
@@ -0,0 +1,123 @@
...
+// ----TID 300 Measurement
+// ------SCOORD. Graphic Type: POLYLINE
+//        (ftp://dicom.nema.org/MEDICAL/dicom/current/output/chtml/part03/sect_C.10.5.html)
+// ----TID 300 Measurement
+// ------SCOORD. Graphic Type: POLYLINE
+//        (ftp://dicom.nema.org/MEDICAL/dicom/current/output/chtml/part03/sect_C.10.5.html)
+//
+// ------ TID 4019 Algorithm Identification:
+//                 AlgorithmName: 'Angle'
+//                 AlgorithmVersion: 'Cornerstone'
```
fedorov: Thinking about this, I would imagine that the concept corresponding to the angle measurement would be sufficient. Can you think of an example where interpretation of the annotation would be ambiguous for a specific angle measurement type defined by the quantity concept?

fedorov: In any case, I would not use Algorithm identification to communicate the semantics of the annotation.

PR #18, opened 2018-10-12
Port over image interactor style for volume-based slice viewing

floryst:
Should be similar to interaction in Slicer's 2D views, with scroll wheel slicing and window/level.

- [ ] Not sure about slicing through an axis with a spacing not equal to 1. Need to test that.
- [ ] Slicing currently does not stop at image bounds.


pieper:
Thanks @floryst - sorry it took a while but when I showed the demo to people they missed the rotate mode so I made it hot swappable.

Issue #30, opened 2018-12-10
Referenced Series Sequence is missing required fields


JamesDarcy616:
SEG generated by dcmjs is missing required fields in Referenced Series Sequence (0x0x00081115)

DICOM part 3 - C.12.2 Common Instance Reference Module
 
* Study Instance UID (0x0020000D) required but missing.
* Referenced Instance Sequence (0x0008114A) required but missing.

Related to issue #4?

JamesAPetts:
Note, the related code is here: https://github.com/dcmjs-org/dcmjs/blob/master/src/derivations.js#L233

I'll look into it.

JamesAPetts:
For future reference for those who find this thread, the `Study Instance UID` is not a required field of of the `Common Instance Reference Module`, but instead a required field of the 1C `Studies Containing Other Referenced Instances Sequence (0008,1200)`, which is only present if the multi-frame references images from a separate study.

Otherwise a PR for the Referenced Instance Sequence is coming in shortly after a bit of testing :+1:

JamesAPetts:
Fix in #31 .

PR #31, opened 2018-12-11
Referenced Series Instance fix

JamesAPetts:
Addresses #30.

- Corrected a bug in `ImageNormalizer` that added `ReferencedInstanceSequence` in an incorrect way.
- Tidied up the `convertToMultiframe` method of `ImageNormalizer`.
- Added a more robust `ReferencedInstanceSequence` generator to Segmentation, should the referenced dataset not have the required information.


pieper:
Thanks @JamesAPetts 👍

PR #33, opened 2019-01-07
add loglevel

NussknackerXXL:
Adding [loglevel](https://www.npmjs.com/package/loglevel) dependency for logging.


pieper:
Thanks!

PR #35, opened 2019-01-30
Seg cornerstone adapter

JamesAPetts:
Adds an adaptor for cornerstonejs to dcmjs which reads/writes seg data to/from cornerstoneTools state.

Updated the create and read seg examples to demonstrate this (also updated the examples to cornerstoneTools v3).


dannyrb:
@JamesAPetts, I think you've accidentally committed some webpack/dev stuff.

JamesAPetts:
@dannyrb We realised, but had to run, so I didn't merge this yet, will get to it tomorrow :)

PR #37, opened 2019-02-01
Move dicomweb.js to source and export the DICOMWEB class

NussknackerXXL:


pieper:
This is okay for now, but why don't we migrate the examples to [dicomweb-client](https://github.com/dcmjs-org/dicomweb-client) and keep dcmjs focused on data representation and manipulation.

PR #39, opened 2019-02-06
removing build folder and updating gitignore file

biharck:
### What
Removing the ‘build’ folder from the actual git repository
### Why
It keeps getting merge conflicts


pieper:
We should be sure to update the examples - currently dcmjs-org.github.io/examples point to ../../build to get their copy of dcmjs, but that's not going to work.

Probably the best would to move them to a new repository, like dcmjs-examples, that uses the unpkg versions of dcmjs and dicomweb-client (even with specific version tags) so they don't keep breaking.

biharck:
@pieper Good catch. I agree with you. I think the examples should be a new repository and keep the lib self-contained.

pieper:
Thanks - for jumping on this!

swederik:
Closing because I merged this in another PR. Thanks!

Issue #42, opened 2019-02-19
Ungraceful errors when undefined values comes to DicomMetaDictionary.denaturalizeValue


biharck:
There is an edge case where an array of null elements comes to **denaturalizeValue** method at https://github.com/dcmjs-org/dcmjs/blob/2865d52fe0d93fc35656882d7a12cca5fdb51c7f/src/DicomMetaDictionary.js#L124

When this situation occurs, this method fails ungraceful:
`Uncaught TypeError: Cannot read property 'constructor' of undefined`
 
Perhaps when there is an array that contains only null values we should throw an exception:
`[undefined, undefined, undefined]`
`throw new Error("naturalValue is completely undefined at DicomMetaDictionary.denaturalizeValue")`

One question not defined yet is, how should behave this method when only some values have undefined values on the array:
`[2.25, undefined, 1.15, undefined, 10.5]`

biharck:
A PR was open in order to fix it. Up to now, it throws an Error if there is at least one undefined value at the array: https://github.com/dcmjs-org/dcmjs/pull/43

PR #43, opened 2019-02-19
adding rule to throw an error when naturalValue is an array with undefined values

biharck:
### What
It was included a new rule to throw an error when there is at least one undefined value under the natural values if it is an array

### Why
When this situation occurs, the `denaturalizeValue` method fails ungraceful:
`Uncaught TypeError: Cannot read property 'constructor' of undefined`

This PR is related to issue https://github.com/dcmjs-org/dcmjs/issues/42

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -120,6 +120,11 @@ class DicomMetaDictionary {
         let value = naturalValue;
         if (!Array.isArray(value)) {
             value = [value];
+        }else{
+          const thereIsUndefinedValues = naturalValue.filter(item => item !== undefined).length > 0;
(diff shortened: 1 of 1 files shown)


src/DicomMetaDictionary.js
```diff
@@ -121,7 +121,7 @@ class DicomMetaDictionary {
         if (!Array.isArray(value)) {
             value = [value];
         }else{
-          const thereIsUndefinedValues = naturalValue.filter(item => item !== undefined).length > 0;
+          const thereIsUndefinedValues = naturalValue.some(item => item === undefined);
```
pieper: Nitpick: should be `thereAreUndefinedValues`

PR #44, opened 2019-02-19
Dicom microscopy adapter

biharck:
### What
A set of changes in order to create an adapter for microscopy viewer.

### Why
There is a necessity to come up with a set of adapter for microscopy. This one is the very beginning to have the first report from a Polyline model


src/adapters/DICOMMicroscopyViewer/MeasurementReport.js
```diff
@@ -0,0 +1,182 @@
...
+      measurementsByGraphicType[graphicType].push(roi.scoord);
+    });
+
+    // For each measurement, get the utility arguments using the adapter, and create TID300 Measurement
+    // Group these TID300 Measurements into a TID1501 Measurement Group (for each graphicType)
+    // Use TID1500MeasurementReport utility to create a single report from the created groups
+    // return report;
+
+    // TODO: Find a way to get this from the measurement itself
+    const ReferencedSOPSequence = {
```
swederik: You can remove this since it isn't used for SCOORD3D.

biharck: done. Thank you!

PR #48, opened 2019-02-28
Add ignoreErrors flag to readFile() to allow partial read

jmhmd:
This adds a flag to readFile() to let the parser return partial results in the event of a read error due to a damaged file, for example a "RangeError: Offset is outside the bounds of the DataView" sort of thing. Should be analogous to dcmtk's dcmdump --ignore-errors flag.

Not sure if allowing more permissive parsing like this is attractive to others, feel free to discuss or suggest other solutions.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -30,17 +30,25 @@ const encapsulatedSyntaxes = [
 ];
 
 class DicomMessage {
-    static read(bufferStream, syntax) {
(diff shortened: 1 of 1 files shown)


pieper:
Good point - seems reasonable to me.  @swederik or @hackermd any thoughts?

swederik:
Yeah this is a good idea. PR looks good to me.

pieper:
Thanks @jmhmd 👍

PR #50, opened 2019-03-06
Reorganization for server

emelalkim:
I moved the message from dicomweb-client as we discussed with Steve. However, multipartdecode was having problems with separating a multipart file I downloaded from dcm4chee (Steve's instance). It was partitioning ok but the dcmjs readFile wasn't finding the DICM it was looking for on the files after the first and rejecting the file saying it is not a dicom. I fixed it by finding 2 consequent crlfs after the boundary. The standard says that if the "preamble" area is not used, the entity headers must be followed by TWO CRLFs (from https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). I'm not sure it it will cover everything, any ideas @swederik


pieper:
Thanks @emelalkim!

PR #62, opened 2019-05-10
Implementation of Structured Reporting

hackermd:
This PR implements structured reporting analogous to the *pydicom* implementation by exposing an intuitive, object-oriented interface for creation of SR documents based on template TID 1500 Measurement Report.

Commit 678543c introduces the `sr` sub-package with the intent to replace `adapters`. The adapter approach has several shortcomings in my opinion. First, the implementation is incomplete and generated SR documents are not standard compliant (several required attributes are missing, etc.). Second, the library would ideally not be aware of other tools, which in turn depend on the library.

The constructors of classes for DICOM templates, content items and value types defined in `sr.templates`, `sr.contentItems` and `sr.valueTypes`, respectively, perform many checks on the `options` object to ensure arguments have the correct type and appropriate values. A lot of these checks could be further abstracted using JSON schemas. However, this would add further package dependencies. Therefore, I would like to defer this decision.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "dcmjs",
-  "version": "0.4.6",
+  "version": "0.5.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "7.0.0"
       }
     },
     "@babel/core": {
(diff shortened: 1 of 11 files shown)


pieper:
For reference, here's the the pydicom discussion thread: https://github.com/pydicom/pydicom/pull/824

I like the idea of having the js and python libraries in sync as much as possible and I wonder how best to maintain this going forward.  @hackermd how much of this code is autogenerated and shouldn't the converter script be maintained so that it's easy to update both codebases in the future?

Also I think you meant to say this would replace the `utilities` not the `adapters` - we still need the code to take, for example, a cornerstoneTools length measurement and populate an SR from it.  But it is true that the `adapters` form another layer and could be factored out into another library or set of libraries.

hackermd:
> How much of this code is autogenerated and shouldn't the converter script be maintained so that it's easy to update both codebases in the future?

None of the code is currently autogenerated. This was all good old manual coding.

After initial discussions with @fedorov at the project week, I refrained form auto-generating code. The main reasons were

1. we wanted to implement only a subset of optional content items either because we didn't understand them or see value in implementing them or because we wanted to avoid duplicated of information across multiple locations in a report document, e.g. header versus observation context)
2. we wanted to abstract the complexity of the templates by providing an intuitive object-orientated interface that's safe for human consumption

Further, I wanted to avoid autogenerating constructor arguments from code meanings, since this approach comes with all sorts of problems (see corresponding pydicom discussion thread). Some of the meanings are very long and would thus also not represent ideal names for constructor arguments.

Since the standard is supposed to be rather stable and not change very frequently, I don't think it's a problem that code is generated and maintained manually.

hackermd:
It would be valuable, however, to have autogenerated JSON schemas, which we could use to validate generated SR datasets (in *naturalized* form and/or *raw* DICOM JSON).

pieper:
Markus and I reviewed the code together.  We want to talk with @swederik and to look at how this would be used, for example, converting the crowds-cure data.

Another thing to look at is integrating the document.js code here with the [derivation classes](https://github.com/dcmjs-org/dcmjs/tree/master/src/derivations).  Also we want to clean out some of the hex numbers and use dictionary names instead.

hackermd:
@pieper have you had a change to look at this?

pieper:
@swederik and I discussed briefly but didn't have time to see what it would look like in practice.  This is going to come up again soon when we add SR (and SEG) support to the react version of OHIF.

swederik:
Sorry for the long delay @hackermd! This looks really good and I see no reason why we shouldn't merge it and start to remove the `utilities` with this instead. I also agree we should push `adapters` out into some other repository. We can update the adapters to start using this `sr` module.

hackermd:
@swederik @pieper I think as long as you feel comfortable about the object-oriented interface for construction of SR template and content item classes, we should be good. We can still change implementation details later.

PR #71, opened 2019-09-02
Cornerstone Tools 4.X Segmentation Adapters

JamesAPetts:
Built against `cornerstone-tools@4.0.0-beta3`.

- Working read for non-overlapping segments (overlapping are outside the scope of the current project that fueled this PR, for now.)
- Working write for non-overlapping and overlapping segments (overlapping segments take no additional implementation in the "write" case).
- Updated `createSegmentation` and `displaySegmentation` examples to use `4.0.0-beta3`. The `createSegmentation` example now lets you draw on two labelmaps with 4 different segments per labelmap, so one can test overlapping segmentation export.

Still to do at some point in the future:
- Read in overlapping segments into multiple `cornerstoneTools` `Labelmap3D` objects.
- Set the color of the segments to that inside the DICOM SEG.


pieper:
Cool - is this ready for review and merge?

JamesAPetts:
Hey, you can feel free to review it, but I think we should hold off on a merge until `cornerstoneTools` 4.0 is officially out.

pieper:
I took a quick look and like it. There's a lot there though, so I feel the need to kick the tires before I could say much.  No particular rush.

JamesAPetts:
I've tested this in OHIF with all the same data. Going to merge since the just released [cornerstoneTools 4.0](https://github.com/cornerstonejs/cornerstoneTools/issues/1061) relys on it.

We can fix any issues we may find down the line 👍  :).

PR #79, opened 2019-10-25
bug(json) map number strings to be numbers

### LINKED ISSUE #53: IntegerString (IS) and maybe other VRs should be Number not String
As reported by @emelalkim, IS elements are reported as quoted strings by dcmjs, but dcm4chee reports them as json Numbers.  Looking at the standard dcm4chee is correct and dcmjs should be updated.

```
dcm4chee:
"00200013": {
  "vr": "IS",
  "Value": [
    10
  ]
},
 

dcmjs
"00200013": {
      "vr": "IS",
      "Value": [
        "10"
      ]
    },
```

http://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_F.2.3.html

pieper:
Fixes #53

@emelalkim could you have a look at this PR and see if this will cause any problems on your end?  It should be backward compatible with anyone who called Number themselves.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -10,186 +10,188 @@ const EXPLICIT_BIG_ENDIAN = "1.2.840.10008.1.2.2";
 const singleVRs = ["SQ", "OF", "OW", "OB", "UN", "LT"];
 
 const encapsulatedSyntaxes = [
-    "1.2.840.10008.1.2.4.50",
-    "1.2.840.10008.1.2.4.51",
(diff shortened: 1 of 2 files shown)


galelis:
@pieper I've tested the changes on OHIF and found one other issue. 
 
Already created a PR on top of this one, so you can review and merge it here if you think its a good fix. and we can keep trying to get approval of this PR. 

Small fix: https://github.com/dcmjs-org/dcmjs/pull/80

pieper:
LGTM, I don't foresee any major problems. Thanks!

PR #80, opened 2019-10-29
Fix number val

galelis:
Issue: 
- Now that some of the values are numbers, val.split not always exists.

Proposed fix: 
- test if val is a string first, if not, we just add the value as it will be a number.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -10,186 +10,188 @@ const EXPLICIT_BIG_ENDIAN = "1.2.840.10008.1.2.2";
 const singleVRs = ["SQ", "OF", "OW", "OB", "UN", "LT"];
 
 const encapsulatedSyntaxes = [
-    "1.2.840.10008.1.2.4.50",
-    "1.2.840.10008.1.2.4.51",
-    "1.2.840.10008.1.2.4.57",
-    "1.2.840.10008.1.2.4.70",
-    "1.2.840.10008.1.2.4.80",
-    "1.2.840.10008.1.2.4.81",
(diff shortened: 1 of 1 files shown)


gbrackley:
We are seeing a regression with release v0.8.1 for tags that are numeric with a multiplicity of greater than 1.  The multiple values are trying to be parsed as a single number which results in a NaN value. A specific value that is failing for us is a ImagePositionPatient (0020,0032) DS value from a CT scan.

Going back to the v0.8.0 release works around this issue. Sorry for the low quality issue reporting.

https://dicom.innolitics.com/ciods/ct-image/image-plane/00200032

pieper:
Thanks for reporting the issue @gbrackley  - that sounds like a pretty important regression.  Do you have a proposed fix?  maybe we just split before converting to numbers? 

@galelis are you able to replicate and suggest a fix?

galelis:
> @galelis are you able to replicate and suggest a fix?
@pieper I will try it today and comment here once I get any updates.

galelis:
@pieper I've checked the code, checked the details into the issue reported and figured out that my fix was not ideal. 

the parse of the DecimalString into a Number only works if we have multiplicity of 1, any greater than one will become a NaN as reported, as it will contain the backslashes and we need to split if before convert each number. 

I will check it again tomorrow, talk with you or James and we can think on a fix for this issue.

galelis:
@pieper and @gbrackley 
I have create a new PR in order to fix the reported issue related with DS (PR #82), few free to review or giving me feedbacks.

I have tested with James help, using segmentation that also uses ImagePatientPosition.

PR #82, opened 2019-11-13
Fix issues on v0.8.1

galelis:
### Issues 
1. Decimal String with multiplicity greater than 1 should be splitted as an array before transcode it as Number
2. We should filter tools that are not supported by Cornerstone adapter measurement report

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -10,188 +10,189 @@ const EXPLICIT_BIG_ENDIAN = "1.2.840.10008.1.2.2";
 const singleVRs = ["SQ", "OF", "OW", "OB", "UN", "LT"];
 
 const encapsulatedSyntaxes = [
-  "1.2.840.10008.1.2.4.50",
-  "1.2.840.10008.1.2.4.51",
-  "1.2.840.10008.1.2.4.57",
-  "1.2.840.10008.1.2.4.70",
-  "1.2.840.10008.1.2.4.80",
-  "1.2.840.10008.1.2.4.81",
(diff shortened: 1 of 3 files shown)


JamesAPetts:
This looks like the right solution, won't you need to check this logic in DicomMessage.js though: 

https://github.com/dcmjs-org/dcmjs/blob/1e10ca74c284e0679189fb87dd872455dcc3aaa5/src/DicomMessage.js#L184

We'll need to check on fallback if the value is an array, and set the values rather than push to them, or we end up with a 2D dimensional array by accident.

i.e. `Array.isArray(val) ? values = val : values.push(val);`

I think that'll be correct if you can test.

pieper:
@galelis can your comment on the issue @JamesAPetts raised?  Would be great to resolve this.

galelis:
> This looks like the right solution, won't you need to check this logic in DicomMessage.js though:
> 
> https://github.com/dcmjs-org/dcmjs/blob/1e10ca74c284e0679189fb87dd872455dcc3aaa5/src/DicomMessage.js#L184
> 
> We'll need to check on fallback if the value is an array, and set the values rather than push to them, or we end up with a 2D dimensional array by accident.
> 
> i.e. `Array.isArray(val) ? values = val : values.push(val);`
> 
> I think that'll be correct if you can test.

@JamesAPetts 
I have update the PR. 
I agree with the change, I also think it will be safer, but I was unable to better test this, as none of the studies I have opened had data to pass into that branch.

JamesAPetts:
This looks correct, and it at least not broken like it is now. Lets get this in and see if any other issues arise.

PR #83, opened 2019-12-02
IS fix

JamesAPetts:
This PR fixes decoding of IS with a multiplicity > 1, similar to #82 .


JamesAPetts:
@swederik , will this solve the issues you were having, (were they related to multiplicity?).

PR #92, opened 2020-02-23
fix reading encoded strings

vsaase:
this is a proposal, may need some testing
the standard DICOM encoding is ISO-IR 6, which I think is something like ASCII. 
More common I think is ISO-IR 100 when not otherwise specified. 
At the moment I only check for UTF-8, known as ISO-IR 192 in the DICOM standard. There may be some other encodings in common use?

### DIFF
--- a/src/BufferStream.js
+++ b/src/BufferStream.js
@@ -62,6 +62,23 @@ class BufferStream {
         this.offset = 0;
         this.isLittleEndian = littleEndian || false;
         this.size = 0;
+
+        this.setEncoding("iso-ir-100");
+    }
+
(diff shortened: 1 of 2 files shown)


pieper:
@vsaase thanks for the contribution 👍 

I'll admit that text encodings is not my area of expertise.  Would it be possible to make some test examples?

I recently added some small amount of testing and an https://github.com/dcmjs-org/data repo to store bulk data as release assets.  Probably we'll continue to evolve the testing architecture but at least for now this gives us at least a little structure.  (Suggestions welcome!).

PR #94, opened 2020-02-24
Fix parsing of WADO-RS headers.

JamesAPetts:
WADO-RS headers often come back with type 2 DICOM tags that have no Value entry. This currently breaks the naturalizer. This fixes that by skipping over these entries.

I'd argue that its more "natural" to not add empty type 2 tags to the "naturalized" (i.e. human readable datasets), but we can add them in later if removing them is deemed unnatural.

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -82,7 +82,7 @@ class DicomMetaDictionary {
     // - single element lists are replaced by their first element
     // - object member names are dictionary, not group/element tag
     static naturalizeDataset(dataset) {
-        var naturalDataset = {
+        const naturalDataset = {
             _vrMap: {}
         };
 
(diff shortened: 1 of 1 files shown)


pieper:
Shouldn't we preserve them as empty.  The standard [says this carries some meaning](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_7.4.3.html).  And in general we should be able to go losslessly to/from natural form.

JamesAPetts:
I was waiting for your comment hehe. Hmm, this makes sense.

I guess we should add them explicitly as `null`. Which will indicate they aren't an error, but are inherently empty.

JamesAPetts:
Do any type 2 SQ exist? I'm not sure if there are any in the standard, they would pose an interesting question as to their value (empty array or null).

JamesAPetts:
http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_5.2.html seems to suggest type 2 SQ exist.

Hmm, I guess we are at the mercy of deciding what happens here. I would think they would still be `null` to suggest the tag is empty, as single entries are replaced by that object.

A lot of logic would currently check if the SQ on the naturalized set is an array or not, and niavely assume its length is non-zero if its an array, which may or may not break some application's assumptions.

pieper:
Yes, I would try to handle them even if we don't have a current example.

Regarding where in the code to deal with it, I'd argue it should be in the Normalizing step, not the Naturalizing.  The Normalizers should be the place where all the special cases are handled, logged, and put into a form so that later parts of the code can only handle actual semantics, not violations of the standard.  The Naturalizing step should just put these into a form that is convenient to program against.

JamesAPetts:
@pieper , I made the changes to add `null` entries, but I'll wait for your opinion on the above 👍 .

JamesAPetts:
> The Naturalizing step should just put these into a form that is convenient to program against.

Sorry I should have put more context in the PR. I'm just naturalizing DICOM JSON already returned from the PACS, so we can use them for WADO-RS headers also. The DICOM headers return type 2's without a `Value`.

In the general case I agree the object returned from `DicomMessage.readFile(arrayBuffer);` Should contain T2s, but the DICOM JSON they made should be in the same form as the WADO-RS headers I'm requesting if they are standard compliant.

pieper:
@JamesAPetts do you think we should have a normalizer step for metadata returned via QIDO and any other DICOM JSON?  Makes some sense that the same logic should apply.

> Was my motivation for removing them in the first place, as having empty keys to check is a bit tedious,

Perhaps it's time to do a high-level code review.  Some of the existing naming conventions could be revisited, and now that we've had a chance to write some sizeable amounts of code we could probably better nail down how to make the code that consumes the library as elegant as possible.

For now though I would be fine with this PR as-is to keep things moving.

JamesAPetts:
>  do you think we should have a normalizer step for metadata returned via QIDO and any other DICOM JSON? Makes some sense that the same logic should apply.

That's an interesting idea, I guess it makes sense to add that at some point.

> Perhaps it's time to do a high-level code review. Some of the existing naming conventions could be revisited, and now that we've had a chance to write some sizeable amounts of code we could probably better nail down how to make the code that consumes the library as elegant as possible.

Yeah this library really could do with some love for a 1.0 at some point. Perhaps after/during more SR work, when its a bit more "feature" complete, we can go through this.

PR #98, opened 2020-03-04
Avoid mutation in static DicomMetaDictionary methods

jody-zeitler:
The static methods `cleanDataset`, `namifyDataset`, and `naturalizeDataset` in DicomMetaDictionary have the side effect of changing the original dataset for sequence tags. `cleanDataset` also modified original values when converting nulls to empty strings.

I added Object.assign and array.map so that the output datasets get their own copies of non-primitives, and the input is safely unchanged.

I encountered this problem when trying to call `DicomMetaDictionary.naturalizeDataset` more than once on the same dataset. It would throw a `TypeError: Cannot read property 'length' of undefined` on the second pass.

Edit: Looks like `naturalizeDataset` was refactored while I was sleeping 😁, so this change now only applies to the clean and namify methods.

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -24,11 +24,11 @@ class DicomMetaDictionary {
     // TODO: if this gets longer it could go in ValueRepresentation.js
     // or in a dedicated class
     static cleanDataset(dataset) {
-        var cleanedDataset = {};
+        const cleanedDataset = {};
         Object.keys(dataset).forEach(tag => {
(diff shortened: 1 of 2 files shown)


pieper:
LGTM - @JamesAPetts, any comments?

JamesAPetts:
Thanks for this, tests especially! Yeah this looks right, good catch. I think most of our "in practice" examples/use cases have only gone one of two directions so far, so this was mostly missed.

pieper:
Thanks!

Issue #84, opened 2019-12-18
toUTF8Array will convert ° into Â°


jssuttles:
When using the write function of DicomDict, writeString uses toUTF8Array. This function takes in the degree symbol and spits it out as 2 bytes. Is there something that I'm missing? I fixed this in my implementation by changing it to something like Buffer.from(str, 'binary') for Node.js.

pieper:
Hi - Can you provide test code to reproduce the issue?

jssuttles:
I've been building out reports with DICOM SR. So, in my specific case, the problem occurs in the ContentSequence of a DICOM SR. However, for example, let's say that you put a ° in the series description, as per the attached example DICOM file. (Github doesn't take DCM files, so it's a TXT.)

[IM-0001-0001.txt](https://github.com/dcmjs-org/dcmjs/files/3980284/IM-0001-0001.txt)

```javascript
var fs = require('fs');
var dcmjs = require('.');
var buf = fs.readFileSync('./IM-0001-0001.txt');
var dicomDict = dcmjs.data.DicomMessage.readFile(buf.buffer);
var newBuf = dicomDict.write();
fs.writeFileSync('./IM-0001-0001-copy.txt', Buffer.from(newBuf));
```

If you open both files in https://rawgit.com/cornerstonejs/dicomParser/master/examples/dumpWithDataDictionary/index.html, you'll notice that SeriesDescription is said to contain binary data. If you open the console and paste in `dataSet.string('x0008103e')`, you'll see that the original has a `°` and the copy has `Â°`. 
That it is said to contain binary data seems to indicate to me that it's somehow not legally DICOM to include a non-ASCII character in a string. Does that sound right? Is there another way I'm supposed to handle degree symbols? (HOROS doesn't seem to parse it correctly in the ContentSequence tag but does seem to parse it in the series description.)

pieper:
Hi @jssuttles - Sorry I haven't had time to look into this.  It sounds like an encoding issue that hasn't gotten much attention in dcmjs.  If you can suggest a solution that would be great.

jssuttles:
The way that I'm currently handling it is to convert to latin1 instead of utf8.

pieper:
Would it be suitable for a pull request?

jssuttles:
As it is, it isn't suitable. I'm using Node Buffers which have the ability to convert to latin1 built in.
`Buffer.from(value, 'binary')`
The following shows something that does what looks like the equivalent for browsers:
[...]

PR #107, opened 2020-03-11
#84 convert to binary instead of to utf8

jssuttles:
I tested this with the file that I gave, and it seemed to work.


pieper:
Thanks for starting this.  Since neither of us is sure if this is the overall correct answer I'm going to put this on the back burner.  To move this forward I think we need to cross-check with other implementations on a variety of datasets as discussed in #84.

Issue #115, opened 2020-04-07
PixelRepresentation tag lost when saving an image 


JoosaT:
When DICOM image is read, written, and reread, ''PixelRepresentation' tag is lost. Produced image is otherwise valid and similar to the original image.

```
let dicomDict = dcmjs.data.DicomMessage.readFile(arrayBuffer);
let dataset = dcmjs.data.DicomMetaDictionary.naturalizeDataset(dicomDict.dict);
dataset._meta = dcmjs.data.DicomMetaDictionary.namifyDataset(dicomDict.meta)
console.log(dataset.PixelRepresentation) // prints 0
dicomDict = dcmjs.data.datasetToDict(dataset);
let arrayBuffer2 = dicomDict.write();

let dicomDict2 = dcmjs.data.DicomMessage.readFile(arrayBuffer2);
let dataset2 = dcmjs.data.DicomMetaDictionary.naturalizeDataset(dicomDict2.dict);
dataset2._meta = dcmjs.data.DicomMetaDictionary.namifyDataset(dicomDict2.meta)
console.log(dataset2.PixelRepresentation) // prints undefined
```

JoosaT:
Simple workaround:

```
let dicomDict = dcmjs.data.DicomMessage.readFile(arrayBuffer);
const pixelRepresentationCode = "00280103"
const pixelRepresentation = dicomDict.dict[pixelRepresentationCode]
let dataset = dcmjs.data.DicomMetaDictionary.naturalizeDataset(dicomDict.dict);
dataset._meta = dcmjs.data.DicomMetaDictionary.namifyDataset(dicomDict.meta)
console.log(dataset.PixelRepresentation) // prints 0
dicomDict = dcmjs.data.datasetToDict(dataset);
dicomDict.dict[pixelRepresentationCode] = pixelRepresentation;
let arrayBuffer2 = dicomDict.write();

let dicomDict2 = dcmjs.data.DicomMessage.readFile(arrayBuffer2);
let dataset2 = dcmjs.data.DicomMetaDictionary.naturalizeDataset(dicomDict2.dict);
dataset2._meta = dcmjs.data.DicomMetaDictionary.namifyDataset(dicomDict2.meta)
console.log(dataset2.PixelRepresentation) // prints 0
```

pieper:
Hi @JoosaT  - I haven't had time to look into this yet, but we should definitely fix this - thanks for the report 👍 

[...]

PR #119, opened 2020-05-06
Adding null verification when denaturalizing the dataset values

andrebot:
After the update done by PR [116](https://github.com/dcmjs-org/dcmjs/pull/116) to fix the issue #115 DCMJS can't parse null values which he sets himself when naturalizing the dataset value at [DicomMetaDictionary#L107](https://github.com/dcmjs-org/dcmjs/blob/master/src/DicomMetaDictionary.js#L107) causing an error to be thrown at [DicomMetaDictionary#L160](https://github.com/JoosaT/dcmjs/blob/9baed8983554d1583cc802c5d6d3dad0e02426bb/src/DicomMetaDictionary.js#L160).

This PR only adds an extra validation to the dataValue, skipping the parsing if the value is either `undefined` or `null`

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -170,7 +170,7 @@ class DicomMetaDictionary {
             var entry = DicomMetaDictionary.nameMap[name];
             if (entry) {
                 let dataValue = dataset[naturalName];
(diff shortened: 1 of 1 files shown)


JamesAPetts:
LGTM, good catch.

PR #121, opened 2020-05-20
Bug: Considering type 2 tags when denaturalizing

andrebot:
When denaturalizing tags we are not considering type 2 tags with null values and removing then from the dataset, which is wrong. This PR will consider null values as correct and jump all parsing and directly return the tag value as null.


JamesAPetts:
Makes sense, good catch.

Issue #125, opened 2020-05-28
dcmjs.data.DicomMessage.readFile() fails for certain dicom SR


ChrisWoodlief:
Hi, we have been using the dcmjs.data.DicomMessage.readFile function on Dicom SRs. We have found a case where other dicom parsing solutions like [Dicom Parser](https://rawgit.com/cornerstonejs/dicomParser/master/examples/dumpWithDataDictionary/index.html) and RadiAnt DICOM Viewer are able to parse the file correctly. But dcmjs is having an issue. Here is how we are using DicomMessage.readFile:

<img width="1253" alt="dcmjs stack" src="https://user-images.githubusercontent.com/5483193/83155681-7a5fe200-a0cf-11ea-8134-3ab02cd9fd04.png">

Here is the file which is causing the issues:
https://medicom-dicoms.s3.amazonaws.com/publiclyposted/dcmjs_parsing_issue_SR

pieper:
Thanks for reporting the issue and supplying the data.  I took a quick look but don't really have time to dig in right now.  Did you try stepping through in the debugger?  If you can host a little site off of github that serves the file and tries to read it with dcmjs it would be easier to dig into.

ChrisWoodlief:
Hi @pieper !

We have tried stepping through in a debugger but once we were really deep into the dcmjs library it became a bit harder to understand the root cause of the issue. From what we have seen it seems like the reading of the tag 0040,A073 is not terminating and the reading goes past the end of the file.

Here is an off site example we made linked up to the file posted above.
https://jsfiddle.net/ds4vrq2g/2/

pieper:
Hmm, yes, looks like maybe it's not handling nested sequences well.  You said you have similar SR files that can be read?  Did you see any obvious differences?   Maybe you can share a similar one that does work.

I inherited this code and it's worked well with a whole lot of datasets and I would have thought that all nested sequences would be basically the same, but I guess there's something unique about this one.  I see a couple sequences with no content so maybe that's not accounted for correctly.

jssuttles:
Hi @pieper, I work together with @ChrisWoodlief.
To be able to post this DICOM here, we needed to anonymize it. When we ran it through a DICOM Editor tool (https://www.dvtk.org/dicom/editor/), the tool added a few characters to that area. Afterwards, when we tried to read the file with dcmjs, it worked. To produce the anonymized dicom as posted, we edited the DICOM as text making sure to alter just the visible characters.

[...]

PR #126, opened 2020-06-01
#125 resolves issue with nested SQ items of empty length

jssuttles:
test at https://jsfiddle.net/24g1rcvb/


jssuttles:
changed to https://jsfiddle.net/xvy8tsr5/ because I forgot to alter comments

PR #135, opened 2020-07-08
Cornerstone findings

JamesAPetts:
If a cornerstone measurement has `findings` or `findingSites`, encode these in the SR. If an SR has `Finding`s or `Finding Site`s, attach these to the cornerstone measurements when the cornerstone adapter is used.

There are also generic `Finding Site` groups added to the TID300 class, that allow you to pass in a `findingSites` array of the form:

```
[
  {
    CodeValue: "SAMPLE FINDING SITE",
    CodingSchemeDesignator: "99dcmjs",
    CodeMeaning: "Sample Finding Site"
  },
  // More entries
]

```

Which will build appropriate `Finding Site` content items, similar to the already existing functionaly for `Findings`.

### DIFF
--- a/examples/cornerstoneDICOMSR/index.html
+++ b/examples/cornerstoneDICOMSR/index.html
@@ -89,7 +89,6 @@
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
         <script src="/js/initCornerstone.js"></script>
-
         <script src="/js/dcmjs.js"></script>
 
         <script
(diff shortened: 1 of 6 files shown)


fedorov:
@JamesAPetts note that "Finding" VM is 1, while "Finding Site" is 1-n. It is probably not a good idea to have multiple findings, if I understand the meaning and context correctly.

http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1410

JamesAPetts:
@fedorov Good call out, I'll make the findings a singular object rather than an array of objects.

JamesAPetts:
Changes made. I'll merge this for now, if there is any other feedback we can look to harden this at a further date.

PR #147, opened 2020-09-07
Fix multiframe read/write

jmhmd:
- Use offset table for reading fragments if present, otherwise, loop until end of stream/sequence

Issue I was seeing was with a multiframe US instance with no offset table (length == 0), and dcmjs would then assume there were no frames. This changes that behavior to go ahead and try to parse the pixel data sequence, which is what dicomParser does.

- Add option to not fragment encapsulated pixel data (use fragments equal to frame size)

Discussion with others on the open source imaging zoom call seemed to suggest that pixel data fragmentation is partly an artifact of older machines that could not buffer entire large instances in memory. This adds an option not to fragment when writing encapsulated pixel data, which in my brief tests seems to be a lot faster for large cine instances.

### DIFF
--- a/src/DicomDict.js
+++ b/src/DicomDict.js
@@ -17,7 +17,7 @@ class DicomDict {
         }
     }
 
-    write() {
+    write(writeOptions = {}) {
         var metaSyntax = EXPLICIT_LITTLE_ENDIAN;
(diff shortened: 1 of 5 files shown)


pieper (approved):
Thanks for the fixes 👍

PR #148, opened 2020-09-07
Add pattern matching for running one test from command line

jmhmd:
This lets you easily run just a subset of the tests, using a simple pattern:

`npm run test [pattern]` will only run tests that contain `[pattern]` in the name, not case sensitive. Helpful for development.


pieper (approved):
👍

PR #151, opened 2020-10-06
Fix rotation issues for different segs.

JamesAPetts:
RE: https://github.com/OHIF/Viewers/issues/2003

Fixes SEG mapping issues for the cornerstone segmentation adapter, when the data is rotated relative to the source data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "@babel/polyfill": "^7.8.3",
     "@babel/runtime": "^7.8.4",
     "loglevelnext": "^3.0.1",
-    "ndarray": "^1.0.19"
+    "ndarray": "^1.0.19",
+    "gl-matrix": "^3.1.0"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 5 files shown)


fedorov:
@pieper let's not rush this into the IDC MVP fork this week (if it gets merged) - we don't have any data included in MVP that would benefit from this fix, and we need to test it first.

pieper:
@fedorov I'd rather integrate this so that we can test it and fix it in OHIF if needed.  Otherwise we'd need to start cherry-picking commits and would have a complex divergence in our fork.

fedorov:
Ok, now that it's merged - let's start testing it. Can you let me know when the sandbox is updated?

JamesAPetts:
> Ok, now that it's merged - let's start testing it.

This is just the dcmjs implementation. I need to merge this to OHIF, and then update the IDC fork. I'll ping you two when its done 👍 .

Issue #144, opened 2020-08-13
_image.data is undefined


KhyatiMehta3:
Hello! 
I'm trying to write a segmentation drawing extension tool over OHIF Viewer. 
I'm referring to [this](https://dcmjs.netlify.app/createsegmentation/) example, to get the Part10 dcmjs representation of the segmentation.
In the example [this image promise](https://github.com/dcmjs-org/dcmjs/blob/7bea0d9524582c8551a5cdc0e82e3badb69b8f0c/examples/createSegmentation/index.html#L173) returns the image without any issues and creates the segmentation, as shown below: 
![seg_issue1](https://user-images.githubusercontent.com/25249564/90092031-814cd700-dd45-11ea-8e83-90f7b1fece93.png)
But in the viewer, the same image Promise returns an image without the property "data" in it, which is causing the mentioned error. Shown below: 
![seg_issue2](https://user-images.githubusercontent.com/25249564/90092049-8ad63f00-dd45-11ea-8287-6c6442f48ebd.png)
![Screenshot at 2020-08-13 08-29-54](https://user-images.githubusercontent.com/25249564/90092064-93c71080-dd45-11ea-9e62-6ed39b347a4b.png)
I wasn't sure if the question is more suited under cornerstone tools repository, so I just put it up here because the error is coming under dcmjs script.
Can anyone help me resolve this issue?

pieper:
Hi @KhyatiMehta3  - it's hard to know from your screenshots.  Are you planning to make your implementation available?  It would be great if you could do that, and start by making a pull request so that others can try your code and maybe help debug it.

It can be tricky to work on features like this that require possible changes to both dcmjs and ohif, but it's possible to set up an environment for testing (I think there are instructions in the ohif docs for that).

KhyatiMehta3:
Hi @pieper thank you for your response! I'll send a PR. and refer this issue there as well.

sinanmb:
Hi @helghast79 , @KhyatiMehta3 , I'm facing the same problem, have you found a way to address it?

pieper:
I'm not sure myself, that sounds like a cornerstone issue with the ways the segmentation tool is implemented.

Segmentation display is working in OHIF with cornerstone and vtk.js, so maybe looking at that implementation will help.  Segmentation creation is not yet in OHIF, just in a few demos like the one in dcmjs.

https://viewer.ohif.org/viewer/1.2.826.0.13854362241694438965858641723883466450351448

[...]

PR #153, opened 2020-11-15
feat add generate segmentation from dataset

helghast79:
In some cases cornerstone image loaders deprive _image object from **_data_** property, like reported here [#144](https://github.com/dcmjs-org/dcmjs/issues/144). The added function uses underlying dcmjs functionality but allow users to generate this dataset array themselves and use it to create the segmentation blob. 

Dataset array can be filled with `cornerstone.metaData.get('instance', imageId)` for each image int the series.

**Important:** each element in dataset needs a _meta property with empty array (` _meta=[] `) otherwise segmentation will be saved without any metadata

### DIFF
--- a/src/adapters/Cornerstone/Segmentation_4X.js
+++ b/src/adapters/Cornerstone/Segmentation_4X.js
@@ -21,7 +21,8 @@ import {
 const Segmentation = {
     generateSegmentation,
     generateToolState,
-    fillSegmentation
+    fillSegmentation,
+    generateSegmentationWithDataset
 };
 
 export default Segmentation;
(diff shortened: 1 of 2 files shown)


pieper (commented):
Looks reasonable to me, but I hope @JamesAPetts has a minute to look.

JamesAPetts:
Thanks for this, makes sense to me!

Only things I would suggest is changing the name from `generateSegmentationWithDataset` to `generateSegmentationFromDatasets`. "From" implies input rather than an output which "with" suggests, and making datasets plural also demonstrates that it expects and array of datasets.

Also could you [expose the call through here](
https://github.com/dcmjs-org/dcmjs/blob/master/src/adapters/Cornerstone/Segmentation.js) in the same way as the current two functions?

You don't need to write an implementation for cornerstoneTools 3, but it'd be good to throw an error if they try to call a v3 method that doesn't exist.

P.S. for readers who stumble upon this thread a year from now:
`cornerstone.metaData.get('instance', imageId)` is a feature within OHIF using its metadata provider, and just returns the entire naturalized DICOM jason for that imageId. Its not a "standard" feature of any cornerstone image loader.

helghast79:
totally agree. Done. Also @pieper & @JamesAPetts, do you think it's better to generate _meta property automatically inside this function if not defined or leave it as it is now? Impact on performance is probably negligible but would facilitate the use of the function since the lack of _meta property will not trigger an error but will produce a segmentation file without metadata.

pieper:
It's been a while since I looked, but maybe the `_meta` would be added by `SegmentationDerivation` or a superclass?

helghast79:
Well, yes and I thinks that's the problem if the datasets have no _meta property already defined. Segmentation class extends from DerivedPixels class which in turn extends from DerivedDataset. So, the Segmentation constructor call it's superclass parent constructors where _meta property is added by assigning the input datasets _meta property to it.

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -57,7 +57,8 @@ function generateSegmentationFromDatasets(
     inputLabelmaps3D,
     userOptions = {}
 ) {
-    //on multiframe images, datasets array contains only 1 element but should work as well
+    //make sure _meta property is present in every element of datasets with at least an empty array
+    datasets = datasets.map(ds => ({ ...ds, _meta: ds._meta || [] }));
```
pieper, line 61: Hi - thanks for this, I'm trying to get my head back into the details.

Wouldn't we get the same effect more generally if [this line](https://github.com/dcmjs-org/dcmjs/blob/master/src/derivations/DerivedDataset.js#L41) where changed to:

`_meta: this.referencedDataset._meta || []`

helghast79: Hi Pieper,

Yes, much better. Thanks

pieper (approved):
Thanks for the contribution 👍

thewildnath:
Hello! Any update on merging this PR? I am currently trying to use dmcjs to save cornerstone-tools segmentations into a DICOM-SEG file, and am running into the same error that this PR fixes.

pieper:
I haven't looked at this in a while - @swederik will you be using this for cs3D segs?  Maybe @sedghi should have a look make sure it works with the new code.

Issue #159, opened 2021-01-13
writeBytes is crashing the browser when using encapsulated files with lots of frames


dpedranti:
Hi,

I'm doing a write() after a data.DicomMessage.readFile and Chrome will freeze up and crash if I use it on larger Dicom files with encapsulated pixel data and lots of frames.  Works great with other images that are big in size.

The file that failed is 117.3 MB with 120 frames.

The issue seems to happen in the for loop on line 258 here: https://github.com/dcmjs-org/dcmjs/blob/master/src/ValueRepresentation.js

Each iteration of that for loop becomes progressively slower until it stops somewhere in the middle of the 120 frames.

Can you help fix?

Thanks!

pieper:
Thanks for the report.  It would really help if you could provide a way to replicate this issue (i.e. sample data and test code to reproduce).  That way it can be profiled and debugged.

dpedranti:
Sure!  And thank you.  Here's a sample image that has encapsulated pixel data and 120 frames:
https://oculusstoragestaging.blob.core.windows.net/error-testing/IM-0001-0010.dcm?sv=2019-12-12&st=2021-01-11T14%3A29%3A14Z&se=2021-02-11T14%3A29%3A00Z&sr=b&sp=r&sig=94kfZANKkjjd1q4GsgMaDkPG8vSEoHKTwQk6vG3A6s0%3D

Here's the code snippet where the problem occurs.  Passing in the arrayBuffer of the aforementioned dicom file.

`let dicomData = data.DicomMessage.readFile(arrayBuffer)
// I have other logic here but not needed for this test as fails with the same issue regardless
let newFileWriterBuffer = dicomData.write()
`
So it's pretty much just doing a write() on dicomData.  It very quickly freezes up and eventually crashes in my Chrome browser.

WoonchanCho:
This is caused by too many BufferStream's "concat" method call to concatenate fragmented binary stream when dealing with a multiframe image, resulting in too many times of ArrayBuffer creation. This leads to O(n^2) on the memory (& time) complexity of writing DICOM data.

The optimization is needed. I just quickly changed the part of source and am seeing an improvement of the performance. 
On my laptop, it ends within a second.
I will raise a PR soon after performing few more tests.

Thank you.

dpedranti:
Thank you very much!  dcmjs is a great library and this fix will really help when working with multiframe images.

WoonchanCho:
I've just raised a PR : #160.
[...]

PR #160, opened 2021-01-18
fix improves the speed of writing bytes for an encapsulated multiframe image (#159)

WoonchanCho:
Hi, team,

This is to fix the issues #159: https://github.com/dcmjs-org/dcmjs/issues/159
It takes a lot of time (or the app crashes with heap out of bound) when writing bytes for an encapsulated multiframe images.

This is caused by too many BufferStream's "concat" method call to concatenate fragmented binary stream when dealing with a multiframe image, resulting in too many times of ArrayBuffer creation. This leads to O(n^2) on the memory (& time) complexity of writing DICOM data.

I tested with a sample image that has encapsulated pixel data and 120 frames:
https://oculusstoragestaging.blob.core.windows.net/error-testing/IM-0001-0010.dcm?sv=2019-12-12&st=2021-01-11T14%3A29%3A14Z&se=2021-02-11T14%3A29%3A00Z&sr=b&sp=r&sig=94kfZANKkjjd1q4GsgMaDkPG8vSEoHKTwQk6vG3A6s0%3D

Before this patch, it took 1,060 seconds.And after the patch, it only took 0.8 seconds with the same output.

If there is any question or anything to change, feel free to let me know.

Thank you so much

### DIFF
--- a/src/BufferStream.js
+++ b/src/BufferStream.js
@@ -245,12 +245,24 @@ class BufferStream {
     }
 
     concat(stream) {
-        var newbuf = new ArrayBuffer(this.offset + stream.size),
-            int8 = new Uint8Array(newbuf);
-        int8.set(new Uint8Array(this.getBuffer(0, this.offset)));
-        int8.set(new Uint8Array(stream.getBuffer(0, stream.size)), this.offset);
-        this.buffer = newbuf;
-        this.view = new DataView(this.buffer);
+        var available = this.buffer.byteLength - this.offset;
+        if (stream.size > available) {
+            let newbuf = new ArrayBuffer(this.offset + stream.size);
+            let int8 = new Uint8Array(newbuf);
+            int8.set(new Uint8Array(this.getBuffer(0, this.offset)));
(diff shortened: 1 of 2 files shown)


pieper:
This looks good to me but I don't have an easy way to test.  @dpedranti  are you able to test this?  Sounds like an important fix (crash and performance).

dpedranti:
Sure!  I'll test it out tonight and let you know.

dpedranti:
I pulled and linked it to the application I'm working on and it fixed the issue with the image that has encapsulated pixel data and 120 frames.  It did the write in less than one second.

Works great!

WoonchanCho:
@dpedranti, Sounds great. thank you so much for testing it out.
@pieper Appreciate if you merge this if you are good to go.

pieper (approved):
LGTM 👍

pieper:
Thanks for the fix and for testing!

dpedranti:
Thanks for working to fix this!

PR #164, opened 2021-01-29
adds support for segmentations that has SegmentIdentificationSequence…

emelalkim:
… in SharedFunctionalGroupsSequence in Segmentation_3X and Segmentation_4X. Updates Segmentation_3X to support SourceImageSequence with one item and adds support for retrieval of SourceImageSequence from SharedFunctionalGroupsSequence in Segmentation_4X


pieper (changes requested):
Makes sense to me.  I don't have any way to test it but it looks backwards compatible.

If you change the commit message to something have a prefix like `feat(seg)`  it will automatically trigger a minor version release to npm.

Issue #167, opened 2021-02-05
File size keep increasing


WoonchanCho:
Hi Team,
I keep testing this out in various ways to check if this library is in stable status and to help improve this library if any anomaly found.
I made a simple script that repetitively just reads a dicom file (DicomMessage.readFile), writes to buffer, and re-reads from the buffer written again and again to test the result at every cycle is produced constantly.

In the test with the attached file, I found that the file size grows at each cycle. The original file size was 4MB, it grew to 6MB at the first iteration, to 11MB at the next iteration, 20MB, 38MB, 74MB, 146MB and then failed with heap memory issue.

The dcm file used is one that I got from case1 at http://www.dclunie.com/pixelmedimagearchive/upmcdigitalmammotomocollection/index.html
I attached the file here too.
[dcm.zip](https://github.com/dcmjs-org/dcmjs/files/5930809/dcm.zip)

I've not looked into the source code yet, just wondered if this is a known issue?

```
dcmdump output-0.dcm |grep SOPClassUID
(0002,0002) UI =SecondaryCaptureImageStorage            #  26, 1 MediaStorageSOPClassUID
(0008,0016) UI =SecondaryCaptureImageStorage            #  26, 1 SOPClassUID`
```
The following is a script that I used for testing.
```
const fs = require('fs');
const path = require('path');
const dcmjs = require('dcmjs');
const {
  data: { DicomMessage },
} = dcmjs;

const inputPath = process.argv[2];
const outputDir = process.argv[3] || path.resolve(__dirname, 'output');
if (!fs.existsSync(outputDir)) {
  fs.mkdirSync(outputDir);
}

function iterate(inputPath) {
  let buffer = fs.readFileSync(inputPath).buffer;
  let json;

  for (let i = 0; i < 100; ++i) {
    usage();
    json = DicomMessage.readFile(buffer);
    usage();
    buffer = json.write();
    usage();
    const outputPath = path.resolve(outputDir, `output-${i}.dcm`);
    usage();
    fs.writeFileSync(outputPath, new Uint8Array(buffer));
  }
}

function usage() {
  const used = process.memoryUsage().heapUsed / 1024 / 1024;
  console.log(`${Math.round(used * 100) / 100} MB`);
}

iterate(inputPath);
```

pieper:
Hi - thanks for testing this - it does appear there is a memory leak somewhere, and also that something extra is ending up in the file.  I suspect these are two unrelated issues.

[...]

PR #168, opened 2021-02-07
fix: Fix UN & AT VR processing logic (#167)

WoonchanCho:
* Change UnknownValue (UN) to be extended from BinaryRepresentation
* Add the writeTwoUint16s method to BufferStream to handle to process AT
* Optimize the readString method of BufferStream
* Adds test case to check the equality of pre-write and post-write buffer


WoonchanCho:
Hi @pieper,
This is a fix for the issue #167.

As you mentioned earlier, several issues together made the bigger issue.
This is the summary of the issue and fix.

1) The situation: 
  I found that the file size keeps growing when iterating reading and writing images by using the previous output as the input of the next iteration.
  I found that only some of the UN types elements (not all) grew in size at each iteration.
  Each iteration took took much time and even worse, the iteration process failed only at the fifth iteration of 100 iterations with the attached dcm file.[dcm.zip](https://github.com/dcmjs-org/dcmjs/files/5938424/dcm.zip)

2) The cause
  a) file growth : UN type is being handled as a string representation while this should be binary.
  b) Slowness & heap out of expection: There were inefficient codes at handling big string text. 
        ` while (index++ < numOfMulti) {
            var charCode = this.readUint8();
            string += String.fromCharCode(charCode);
        }`
        The above code unnecessary creates 1, 2, 3, ... nth string. 
        This will be okay at small text, but in case of big text, it consumes too many heap memory (garbage collection takes too much time) and makes program fail.

3) Fixes
    a) Change UnknownValue (UN) to be extended from BinaryRepresentation
    b) Add the writeTwoUint16s method to BufferStream to handle to process AT
        => This is a separate issue, I found that the AT handling logic is not correct.
    c) Optimize the readString method of BufferStream
    d) Add test case to check the equality of pre-write and post-write buffer
    e) Add Iterate-test.js into test folder

After this fix, the 100 iteration test with the above file finished in 3 secs with much less heap used.

You can test it too if you want.

Thank you.

src/ValueRepresentation.js
```diff
@@ -955,17 +955,17 @@ class UniversalResource extends StringRepresentation {
...
         this.padByte = "00";
         this.noMultiple = true;
     }
 
-    readBytes(stream, length) {
-        return stream.readString(length);
-    }
+    // readBytes(stream, length) {
+    //     return stream.readString(length);
+    // }
```
pieper, line 968: Better to remove completely rather than leave around dead code.

test/iteration-test.js
```diff
@@ -0,0 +1,43 @@
...
+async function usage(title, i) {
+    console.log(`---------- [${i}] ${title} ----------`)
+    const dict = process.memoryUsage();
+    Object.keys(dict).forEach(key => {
+        const used = dict[key] / 1024 / 1024;
+        console.log(`${key}: ${Math.round(used * 100) / 100} MB`);
+    });
+}
+
+iterate(inputPath);
```
pieper, line 43: Is this used anywhere?  Can it be made into a test?

pieper (changes requested):
Thanks for working on this 👍   The fix looks good to me.

Would it be possible for you to put your file.dcm.zip as a release asset in the data repository and set up the iterate as part of the test suite? https://github.com/dcmjs-org/data  That will help ensure this doesn't pop up again.

pieper:
Thanks - is it ready to merge?

WoonchanCho:
@pieper Yes, you can merge now.

dcm.zip is the one I found on the internet: http://www.dclunie.com/pixelmedimagearchive/upmcdigitalmammotomocollection/index.html
I just emailed the owner of the images asking if one of those files can be used for this open source project.
I will add dcm.zip once he has approved.

Thank you.

WoonchanCho:
@pieper, this is not related to this issue. I have some questions (not an issue all) on the library.
What will be the best place to ask my questions?

pieper:
Thank you @WoonchanCho 👍   

If the data is from David Clunie he's very likely to approve of using it for this purpose.  I talk with him frequently on another project so if you don't hear back from him just ping me and I can ask.

Regarding any questions about his library, I guess the easiest is to file an issue in the repo here.

WoonchanCho:
@pieper, oh, it's small world, I just randomly googled it to find a breast tomo image :)
I will do so unless I've not heard from hi.

Thank you so much!👍

Issue #170, opened 2021-02-11
"require is not defined" issue in the browser


WoonchanCho:
Hi @Punzo and @pieper

The most recent umd version is causing error when being used with a script tag.

<script type="text/javascript" src="https://unpkg.com/dcmjs"></script>
Uncaught ReferenceError: require is not defined
    at Segmentation_4X.js:27
    at dcmjs.js:4
    at dcmjs.js:5

This seems to be caused by a code in Segmentation_4X.js which is added in #155 
const cloneDeep = require("lodash.clonedeep");

As a quick fix, the above can be changed to the import statement.

Or to prevent this from happening again, we might change a rollup setting to allow mixed modules by adding transformMixedEsModules options.
It seems the current rollup version (commonjs plugin) doesn't support the above options.
For this, a rollup version should be upgraded too.

pieper:
Thank you @WoonchanCho yes, if you can suggest a PR to fix this it would be much appreciated!

WoonchanCho:
@pieper thank you so much for merging.

PR #171, opened 2021-02-11
fix "require is not defined" issue in the browser #170

WoonchanCho:
I just changed require to the import statement.

I didn't upgrade rollup this time. 
Rollup seems to have a breaking change in a newer version, but I'm not familiar with the library.


pieper:
Thanks for the fix 👍

PR #176, opened 2021-02-21
fix: checking the length before writing a DecimalString

### LINKED ISSUE #175: DecimalString values that use exponential notation are converted when reading a DICOM
Given a DICOM that contains a DS value with exponential notation, eg.
`DS: [-0.2437435686588, 0.96983969211578, 7.1945578383e-05, 0.00072092906339, 0.00025536940665, -0.9999997019767]`
When I run DicomMessage.readFile and write
Then write should complete without any errors

Currently the write fails with the following error:
Error: Value exceeds max length, vr: DS, value: 0.000071945578383, length: 17
    at DecimalString.writeBytes

Example code:
```
const fs = require("fs");
const dcmjs = require("dcmjs");

const dicom = fs.readFileSync("exponentialNotation.dcm");
const readFile = dcmjs.data.DicomMessage.readFile(dicom.buffer);
const buffer = readFile.write();
fs.writeFileSync("output.dcm", buffer);
```

Example DICOM:
[exponentialNotation.zip](https://github.com/dcmjs-org/dcmjs/files/5994305/exponentialNotation.zip)

dcmjs version: 0.16.6 (I've also tried the latest)
node version: v12.6.2

TimMullender:
Fixes #175 

Attempt to write the value as an exponential if it is too long for the DecimalString VR.

Note this does not address the issue of the format being changed for some values nor does it allow a user to specify the format to use for a value


src/ValueRepresentation.js
```diff
@@ -497,8 +497,18 @@ class DecimalString extends StringRepresentation {
         return ds;
     }
 
+    formatValue(value) {
+        const str = String(value);
+        if (str.length > this.maxLength) {
+            return value.toExponential();
```
pieper, line 503: I wonder if we shouldn't add the fractionDigits parameter here.  From what I can tell 10 would probably be a good number, but maybe the best would be to start with a number like 17 and then decrement it in a loop until the length of the string is less than or equal to 17.  This would give us the most precision for a valid VR but automatically adapt to minus signs for the number or exponent.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toExponential

pieper (changes requested):
This looks great, thanks for the fix 👍 

let me know what you think about the fractionDigits suggestion

src/ValueRepresentation.js
TimMullender: Limiting the number of digits would lead to a loss of precision, my preference would be to fail (in a clear and obvious way) rather than silently lose precision.  If the value is not modified (i.e. it is just read and rewritten) then it should always fit in one format or the other.  If the value has been modified (i.e. set up a user or calculated) then I think write should fail if it is not a valid value rather than truncating it. It puts the onus for making sure the value is correct back with the user/developer.  Does that make sense?

TimMullender: This reminds me it would be nice if the error message contained the tag as well as the value, if I try to put together a separate PR, would it be useful?

pieper: How about start with `toExponential` with no arguments as you suggest, so that we restore the original value if that's where the number came from.

But if the user calculated a number that turns out to be too long for a DS then I don't think that should be an error condition, at least not by default.  Instead I think dicom should be considered 'lossy' with respect to the 64 bit float Number type in js and it's okay to silently round to the best available representation that is valid in the standard.

I wouldn't see it as needed for my use cases, but a user of the library who is concerned about this lossiness could re-read the object and compare the values.  This would be more elegant than throwing an error to my eye.  Or just call the `formatValue` method to learn the precision of the DS version of the number and take whatever action they think is appropriate.

pieper: Here's how this issue is being addressed in highdicom: https://github.com/MGHComputationalPathology/highdicom/pull/57

pieper (approved):

pieper (changes requested):
There was a merge conflict with another change in master.  I've tried to resolve it but need to have another look.

pieper (approved):
Reading through the file it looks like the merge worked but the diff is broken for some reason and it shows the whole file as different even though it's not.

PR #178, opened 2021-02-23
Update dictionary from the DICOM standard documentation

jody-zeitler:
Our de-identification service is encountering tags that are not defined in the dictionary. I did an audit and found around 800 tags that are missing from the dictionary module. I took the [approach of PyDICOM](https://github.com/pydicom/pydicom/blob/8112bb69bfc0423c3a08cb89e7960defbe7237bf/source/generate_dict/generate_dicom_dict.py) to generate the missing tags with a Node.js script that parses the DICOM standard documentation XML and adds new tags to the dictionary module without clobbering the existing format and ordering. GitHub won't render the diff but it looks pretty clean in VSCode + GitLens.


pieper:
Very nice - thank you! 👍

PR #149, opened 2020-09-07
Add ability to parse `untilTag` and to `stopAfterTag` with namified or Group,Element syntax

jmhmd:
Adds `DicomMessage.readFile()` options `untilTag` and `stopAfterTag` that terminate parsing and return the so far parsed data either up to or including the given tag. Adds normalization of namified or group,element type tags.

Use:
```js
dicomData = dcmjs.data.DicomMessage.readFile(part10Buffer, {
    untilTag: 'StudyInstanceUID', // OR
    untilTag: '(0020,000D)', // OR
    stopAfterTag: 'PatientName', // OR
    stopAfterTag: 'x00100010', // etc...
});
```
Group,Element formats parsed:
```
(0028,7017)
[0028,7017]
0028,7017
00287017
x00287017
```

PR #181, opened 2021-03-02
ignorePixelData

Ouwen:
In some situations the user may be only interested in the metadata from a dicom and would like to ignore the values from a pixel data tag.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -122,7 +122,7 @@ class DicomMessage {
         return written;
     }
 
-    static readTag(stream, syntax) {
+    static readTag(stream, syntax, ignorePixelData = false) {
         var implicit = syntax == IMPLICIT_LITTLE_ENDIAN ? true : false,
             isLittleEndian =
                 syntax == IMPLICIT_LITTLE_ENDIAN ||
@@ -135,6 +135,10 @@ class DicomMessage {
         var group = stream.readUint16(),
             element = stream.readUint16(),
             tag = tagFromNumbers(group, element);
+
+        if (ignorePixelData && tag.isPixelDataTag()) {
+            return { tag: tag, vr: null, values: null };
+        }
 
         var length = null,
             vr = null,


swederik:
This might be another option: https://github.com/dcmjs-org/dcmjs/pull/134

pieper:
Yes, I think I like the more general `stopAfterTag` and `untilTag`.  `PixelData` is almost always the last tag, but in case it isn't is there a case for skipping past it?

Ouwen:
I like the #134 API using `untilTag` for arbitrary tags. 

The main use case I have is for users performing partial read of a dicom file over network. A 5kb request will likely read the DICOM metadata header and part of the pixel data. If no pixel data is found more bytes can be requested until some max size or the end of the file. In this case the pixel data byte stream will be cut at some arbitrary length. We would simply want to check for pixelDataTag existence as a stop, without needing to read the values.

In an ideal world, DICOM would have a metadata tag that points to when metadata bytes end. I'm not a dicom expert so I'm not sure if such a tag exists.

The two features for the network partial read use case would require the following:
1.  some flag for the user as to whether the `untilTag` option is triggered
2.  Avoid reading the values from the untilTag which may throw an error for an early stream termination and result in unnecessary work

Regarding 2. I think the `readTag` function would have to be intercepted to avoid doing work below:
https://github.com/dcmjs-org/dcmjs/blob/b17ea9ab8205590d24f55374cf16dcd43c0b7f40/src/DicomMessage.js#L139-L197

Ouwen:
I like the #134 API using `untilTag` for arbitrary tags. 

The main use case I have is for users performing partial read of a dicom file over network. A 5kb request will likely read the DICOM metadata header and part of the pixel data. If no pixel data is found more bytes can be requested until some max size or the end of the file. In this case the pixel data byte stream will be cut at some arbitrary length. We would simply want to check for pixelDataTag existence as a stop, without needing to read the values.

In an ideal world, DICOM would have a metadata tag that points to when metadata bytes end. I'm not a dicom expert so I'm not sure if such a tag exists.

The two features for the network partial read use case would require the following:
1.  some flag for the user as to whether the `untilTag` option is triggered
2.  Avoid reading the values from the untilTag which may throw an error for an early stream termination and result in unnecessary work

Regarding 2. I think the `readTag` function would have to be intercepted to avoid doing work below:
https://github.com/dcmjs-org/dcmjs/blob/b17ea9ab8205590d24f55374cf16dcd43c0b7f40/src/DicomMessage.js#L139-L197

pieper:
That sounds like a reasonable scenario.  Do you think you could integrate that with https://github.com/dcmjs-org/dcmjs/pull/149 to handle that use case?

Ouwen:
Sure happy to help make a PR. Regarding #149, we could add an `ignoreValue` option in addition to the `untilTag`. Then the flag will be existence of the untilTag in the dicom dict.

Ouwen:
Closing due to #134 thanks @whatwilliam @pieper

Issue #162, opened 2021-01-19
Enable a force write option for files with abnormal VRs


ejcurtis:
Hi all, I am a developer on the Dicom Standard Browser. We are using dcmjs to view and edit .dcm files in the browser. At the moment, the only files we can do this with are those that pass all of the VR validation in [this file](https://github.com/dcmjs-org/dcmjs/blob/09cea9d2b1819a937ecdd1b81aec2f51766a2446/src/ValueRepresentation.js#L157). Some of the files we are manipulating have abnormal VR's but we would still like to be able to write them to a file. We were wondering if there could be an option to force the file to write, even if some fields do not match their VR.

For reference: One of the errors we got is shown below
![VR error](https://user-images.githubusercontent.com/53452158/105096446-b1bf3c80-5a6c-11eb-8fea-9ea2e279477f.png)

In this instance, we would like to be able to force that field to write to the file even if the value exceeds its max characters.

I am tagging @johndgeise to give more context.

johndgiese:
Hey @pieper, I hope you're doing well.

In addition to the ability to write an invalid file, we also were hoping validate dcmjs DicomDicts somehow. Can you confirm that neither of these features are present in `dcmjs`? Also, do you think one or both of these features would be appropriate to include within `dcmjs`?  If so, we may put together a PR to add this functionality to the library.

pieper:
👋  @johndgiese and @ejcurtis - yes, I think if it's exposed in an options object with a name like `allowInvalidVRLength` that is false by default we should be fine.  It's always a tough call to support non-standard formats but since you are just trying to work losslessly with existing fields it makes sense.

When you say validating DicomDicts do you mean something like [dciodvfy](http://dclunie.com/dicom3tools/dciodvfy.html)?  Yes, I think that would be very nice.  Either as a part of dcmjs or as another package in dcmjs-org.

johndgiese:
Thanks @pieper , we'll checkout the `allowInvalidVRLength` option. I was actually thinking of just basic VR validation, but something more sophisticated like [http://dclunie.com/dicom3tools/dciodvfy.html](dciodvfy) would be even better. I think for our first version of our DICOM editor we'll skip validation. Perhaps if the feature is widely used we'll consider adding it later.

Thanks for the quick response.

pieper:
Sounds good @johndgiese.  I've often thought we could do a json schema for this but never took it past the concept stage.  Looking forward to seeing your editor someday.

[...]

PR #186, opened 2021-03-04
Enable a force write option for files with abnormal VRs #162

WoonchanCho:
Hi @pieper or maintainers,
This is an improvement to add a force write option as discussed in #162.
@johndgiese 's team is busy doing other priorities, so I worked on this. Hopefully they also like this change.

The decedents of ValueRepresentation has different parameter list. So, I had to align a bit to add write option
Feel free to give me your thought and push if it looks good.

- add allowInvalidVRLength (false by default) in the write options
- add a test case for testing this option


pieper:
Looks good to me.  @johndgiese and @ejcurtis any comments?

johndgiese:
I took a quick look and the changes look good to me too.

pieper:
Thanks @johndgiese 

Okay, I'll merge with a new commit prefix to trigger a minor release.

Issue #184, opened 2021-03-03
checkSEGsOverlapping throwing Uncaught RangeError for examples/displaySegmentation/index.html


gangaramstyle:
After the patch in Segmentation_4X.js to properly compute getSegmentIndex as described in https://github.com/dcmjs-org/dcmjs/issues/177, there are still a few more small issues to get examples/displaySegmentation/index.html to work without errors. All of the fixes, including the getSegmentIndex patch, are described in https://github.com/gangaramstyle/dcmjs/commit/c8a5ba9. @igoroctaviano has created a commit to fix getSegmentIndex with https://github.com/dcmjs-org/dcmjs/pull/183 but after that is patched we'll likely see the error

Start offset -153792576 is outside the bounds of the buffer
  | _loop4 | @ | Segmentation_4X.js:643
-- | -- | -- | --
  | checkSEGsOverlapping | @ | Segmentation_4X.js:634
  | generateToolState$1 | @ | Segmentation_4X.js:355
  | generateToolState$2 | @ | Segmentation.js:66
  | parseSeg | @ | index.html:129
  | (anonymous) | @ | index.html:98
  | load (async) |   |  
  | getAndLoadSeg | @ | index.html:97
  | onclick | @ | index.html:25

I haven't recursed into how exactly to fix this underlying error here, but the error can be bypassed if Segmentation_4X.js generateToolState skipOverlapping is set to true. Unfortunately the client can't set skipOverlapping to true because the client calls generateToolState from Segmentation.js which then calls Segmentation_4X.js generateToolState without passing a parameter from the client to skipOverlapping.

Segmentation_4X.js
```
code for function generateToolState(
    imageIds,
    arrayBuffer,
    metadataProvider,
    skipOverlapping = false
) {
```

Segmentation.js (current)
```
function generateToolState(
    imageIds,
    arrayBuffer,
    metadataProvider,
    cornerstoneToolsVersion = 4
) {
    if (cornerstoneToolsVersion === 4) {
        return Segmentation_4X.generateToolState(
            imageIds,
            arrayBuffer,
            metadataProvider
        );
    }

    if (cornerstoneToolsVersion === 3) {
        return Segmentation_3X.generateToolState(
            imageIds,
            arrayBuffer,
            metadataProvider
        );
    }
...
```

Segmentation.js (proposed)
```
function generateToolState(
    imageIds,
    arrayBuffer,
    metadataProvider,
    **skipOverlapping = false,**
[...]

PR #185, opened 2021-03-03
fix(Segmentation): Cascade flag to skipOverlapping for generateToolState to the client

gangaramstyle:
This PR and the recent PR https://github.com/dcmjs-org/dcmjs/pull/183 by @igoroctaviano are the steps I needed to take to get the examples/displaySegmentation/index.html to work locally without errors. The rationale for the changes for this specific PR are explained in https://github.com/dcmjs-org/dcmjs/issues/184.

I'm not sure if lines 64 and 66 in displaySegmentation/index.html are necessary for the hosted netlify examples but I had to do that to test the example locally. I can remove those two lines if it's confirmed that they're unnecessary and would actually break the hosted example.

PR #187, opened 2021-03-05
Fix example

Punzo:
Fix https://github.com/dcmjs-org/dcmjs/issues/184

@pieper in this PR I fix the example and add additional checks. 

- Please let me know what do you think of [PR inline comments](https://github.com/dcmjs-org/dcmjs/compare/master...Punzo:fixExample?expand=1#diff-077798721d0df020ee8184a2f54d4459dba916e19c86ac0f0aecc4e6ea80bf2fR632-R637), if the inline comments are not clear we can have a videocall or discuss it at the next IDC-OIHF meeting. 
- The data from the example (**Display DICOM Segmentation IOD**) do not satisfy assumption 1. 
- Data that I have from IDC satisfy both Assumption 1 and 2.

@gangaramstyle this include fix from https://github.com/dcmjs-org/dcmjs/pull/185 as well

### DIFF
--- a/examples/cornerstoneDICOMSR/index.html
+++ b/examples/cornerstoneDICOMSR/index.html
@@ -89,7 +89,7 @@
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
         <script src="../js/initCornerstone.js"></script>
-        <script src="../../build/dcmjs.js"></script>
+        <script src="../js/dcmjs.js"></script>
 
         <script
             src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/cornerstoneDICOMSR/index.html
+++ b/examples/cornerstoneDICOMSR/index.html
@@ -88,8 +88,8 @@
         <script src="https://unpkg.com/cornerstone-wado-image-loader@3.0.2/dist/cornerstoneWADOImageLoader.min.js"></script>
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-        <script src="../js/initCornerstone.js"></script>
-        <script src="../js/dcmjs.js"></script>
+        <script src="/js/initCornerstone.js"></script>
+        <script src="/js/dcmjs.js"></script>
 
         <script
             src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/cornerstoneDICOMSR/index.html
+++ b/examples/cornerstoneDICOMSR/index.html
@@ -88,9 +88,8 @@
         <script src="https://unpkg.com/cornerstone-wado-image-loader@3.0.2/dist/cornerstoneWADOImageLoader.min.js"></script>
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-        <script src="/js/initCornerstone.js"></script>
-
-        <script src="/js/dcmjs.js"></script>
+        <script src="../js/initCornerstone.js"></script>
+        <script src="../../build/dcmjs.js"></script>
 
         <script
             src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/createSegmentationRLERoundtrip/index.html
+++ b/examples/createSegmentationRLERoundtrip/index.html
@@ -66,7 +66,7 @@
     <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
     <script src="../js/initCornerstone.js"></script>
-    <script src="../../build/dcmjs.js"></script>
+    <script src="../js/dcmjs.js"></script>
 
     <script
       src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/createSegmentationRLERoundtrip/index.html
+++ b/examples/createSegmentationRLERoundtrip/index.html
@@ -65,9 +65,8 @@
     <script src="https://unpkg.com/cornerstone-wado-image-loader@3.0.2/dist/cornerstoneWADOImageLoader.min.js"></script>
     <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-    <script src="/js/initCornerstone.js"></script>
-
-    <script src="/js/dcmjs.js"></script>
+    <script src="../js/initCornerstone.js"></script>
+    <script src="../../build/dcmjs.js"></script>
 
     <script
       src="https://unpkg.com/react@16/umd/react.development.js"
@@ -205,7 +204,7 @@
         const t0 = performance.now();
 
         const {
-          labelmapBuffer,
+          labelmapBufferArray,
           segMetadata,
           segmentsOnFrame
         } = dcmjs.adapters.Cornerstone.Segmentation.generateToolState(
@@ -220,7 +219,7 @@
 
         setters.labelmap3DByFirstImageId(
           imageIds[0],
-          labelmapBuffer,
+          labelmapBufferArray[0],
           1,
           segMetadata,
           imageIds.length,
--- a/examples/createSegmentationRLERoundtrip/index.html
+++ b/examples/createSegmentationRLERoundtrip/index.html
@@ -65,8 +65,8 @@
     <script src="https://unpkg.com/cornerstone-wado-image-loader@3.0.2/dist/cornerstoneWADOImageLoader.min.js"></script>
     <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-    <script src="../js/initCornerstone.js"></script>
-    <script src="../js/dcmjs.js"></script>
+    <script src="/js/initCornerstone.js"></script>
+    <script src="/js/dcmjs.js"></script>
 
     <script
       src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/displaySegmentation/index.html
+++ b/examples/displaySegmentation/index.html
@@ -61,8 +61,8 @@
         <script src="https://unpkg.com/cornerstone-wado-image-loader/dist/cornerstoneWADOImageLoader.min.js"></script>
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-        <script src="../js/initCornerstone.js"></script>
-        <script src="../js/dcmjs.js"></script>
+        <script src="/js/initCornerstone.js"></script>
+        <script src="/js/dcmjs.js"></script>
 
         <script
             src="https://unpkg.com/react@16/umd/react.development.js"
--- a/examples/displaySegmentation/index.html
+++ b/examples/displaySegmentation/index.html
@@ -62,8 +62,7 @@
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
         <script src="../js/initCornerstone.js"></script>
-
-        <script src="../js/dcmjs.js"></script>
+        <script src="../../build/dcmjs.js"></script>
 
         <script
             src="https://unpkg.com/react@16/umd/react.development.js"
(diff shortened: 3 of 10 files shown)


examples/displaySegmentation/index.html
```diff
@@ -62,8 +62,7 @@ <h1>Display DICOM Segmentation IOD</h1>
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
         <script src="../js/initCornerstone.js"></script>
-
-        <script src="../js/dcmjs.js"></script>
+        <script src="../../build/dcmjs.js"></script>
```
gangaramstyle, line 65: When trying the netlify preview for this PR https://deploy-preview-187--dcmjs.netlify.app/displaysegmentation/ I get the error 
```
Uncaught ReferenceError: dcmjs is not defined
    at parseSeg ((index):128)
    at XMLHttpRequest.<anonymous> ((index):97)
parseSeg @ (index):128
(anonymous) @ (index):97
load (async)
getAndLoadSeg @ (index):96
onclick @ (index):25
```
Not sure if having it work on the netlify preview is important, but I think that `<script src="../js/dcmjs.js"></script>` might fix it given the behavior in https://deploy-preview-185--dcmjs.netlify.app/displaysegmentation/

gangaramstyle: Looking at the build rules in package.json, it looks like when you run npm run build:examples it expands to
`npm run build && npx cpx 'build/**/*.{js,map}' examples/js`
in which case I believe we should be fine to import from ../js/* in the index.html files

other than that, tested it locally and it works great for the displaySegmentation example!

gangaramstyle:
Other than my comment on importing from build vs ../js/* it looks great to me! Thanks @Punzo

examples/displaySegmentation/index.html
Punzo: oh yeah, you are right, I was testing locally. thanks. I will switch back to src="../js/dcmjs.js"

examples/createSegmentationRLERoundtrip/index.html
```diff
@@ -205,7 +204,7 @@ <h1>Create DICOM Segmentation IOD</h1>
         const t0 = performance.now();
 
         const {
-          labelmapBuffer,
+          labelmapBufferArray,
```
pieper, line 207: why `BufferArray`?  Isn't this an `ArrayBuffer`?

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -651,16 +688,36 @@ function checkSEGsOverlapping(
...
-            inPlane = false;
-            break;
+            continue;
         }
 
         const segmentIndex = getSegmentIndex(multiframe, i);
 
+        if (segmentIndex === undefined) {
+            console.warn(
+                "Could not retrieve the segment index, skipping this frame. "
```
pieper, line 707: We should be thinking of ways to send this kind of information back to the app level in a consumable way.  End users shouldn't be needing to look for console.warn messages.

examples/cornerstoneDICOMSR/index.html
```diff
@@ -88,9 +88,8 @@ <h1>
         <script src="https://unpkg.com/cornerstone-wado-image-loader@3.0.2/dist/cornerstoneWADOImageLoader.min.js"></script>
         <script src="https://unpkg.com/hammerjs/hammer.min.js"></script>
 
-        <script src="/js/initCornerstone.js"></script>
-
-        <script src="/js/dcmjs.js"></script>
+        <script src="../js/initCornerstone.js"></script>
+        <script src="../js/dcmjs.js"></script>
```
pieper, line 92: These paths are different between local testing and deployed examples.  Are you sure they should be changed as part of this PR?

pieper (changes requested):
Generally looks good and we want the seg in idc fixed.  Let me know what you think of my comments.

examples/cornerstoneDICOMSR/index.html
gangaramstyle: So I was the one who initially changed these but I wasn't sure if they should be changed from /js/ to ../js/. I did so to make it work locally and then kept it when I realized that the deploy preview (https://deploy-preview-187--dcmjs.netlify.app/) worked with the change to ../js/ as well. But I did note that the currently deployed previews work great with /js/ rather than ../js/.

examples/createSegmentationRLERoundtrip/index.html
Punzo: no now it's an array of buffers (N elements = N of overlapping segmentations). In the case there are no overlapping segmentation it is still an array of 1 element. We could rename the variable names if you think is not clear enough (e.g.: labelMapsBuffer / labelMapsSeriesBuffer/....?)  .

src/adapters/Cornerstone/Segmentation_4X.js
Punzo: I agree, we can talk about this at the next OIHF-IDC meeting. What about we expand the new Logger service implemented by @igoroctaviano to catch also info/log/warn and not only errors? https://github.com/OHIF/Viewers/pull/2283

examples/cornerstoneDICOMSR/index.html
Punzo: I think @gangaramstyle is right, if you go to the current deployed example of the current master branch (e.g.: https://dcmjs.netlify.app/displaysegmentation/) are broken and they return this error:

Uncaught ReferenceError: cornerstoneWADOImageLoader is not defined
    at initCornerstone.js:40

while it works in https://deploy-preview-187--dcmjs.netlify.app/displaysegmentation/ (i.e.: importing by ../js/).

Punzo:
@pieper thanks for the review. I have replied to the comments

examples/cornerstoneDICOMSR/index.html
Punzo: Update: I can't reproduce this, probably I had to clear my web browers cache. I will delete reverse teh commit modifying the import.

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -629,6 +629,43 @@ function checkSEGsOverlapping(
     var groupsLen = PerFrameFunctionalGroupsSequence.length;
     var numberOfSegs = multiframe.SegmentSequence.length;
     var numberOfFrames = imageIds.length;
+    // NOTE:
+    // Assumption1 : numberOfSegs * numberOfFrames = groupsLen,
+    // i.e. for each frame we have a N PerFrameFunctionalGroupsSequence, where N is the numberOfSegs.
+    // Assumption2 : the order of the group sequence is = numberOfFrames of segmentation 1 +  numberOfFrames of segmentation 2 + ... + numberOfFrames of segmentation numberOfSegs
```
igoroctaviano, line 635: Lets move this to jsDocs

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -629,6 +629,43 @@ function checkSEGsOverlapping(
...
+    // -------------------
+    // We could check the ImagePositionPatient and working in 3D coordinates (instead of indexes) and remove the assumptions,
+    // but this would greatly increase the computation time (i.e. we would have to sort the data before running checkSEGsOverlapping).
+
+    if (numberOfSegs * numberOfFrames !== groupsLen) {
+        console.warn(
+            "checkSEGsOverlapping: it not possible performing the overlapping segments check since PerFrameFunctionalGroupsSequence has missing data."
```
igoroctaviano, line 642: PerFrameFunctionalGroupsSequence is type 1, maybe we should just break or define a new message?
Failed to perform overlap of segments: missing PerFrameFunctionalGroupsSequence

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -629,6 +629,43 @@ function checkSEGsOverlapping(
...
+        );
+        return false;
+    }
+
+    const refSegFrame0 =
+        PerFrameFunctionalGroupsSequence[0] &&
+        PerFrameFunctionalGroupsSequence[0].SegmentIdentificationSequence
```
igoroctaviano, line 649: https://github.com/dcmjs-org/dcmjs/blob/9f33cefc3cefcd1a0ce5f2c18628987e5f1e49d5/src/adapters/Cornerstone/Segmentation_4X.js#L894 

I have used this method to get the ref segment number. Maybe we should use the same logic here? there I'm considering the shared group sequence as a fallback if missing per frame.

src/adapters/Cornerstone/Segmentation_4X.js
```diff
@@ -629,6 +629,43 @@ function checkSEGsOverlapping(
...
+            ? PerFrameFunctionalGroupsSequence[0].SegmentIdentificationSequence
+                  .ReferencedSegmentNumber
+            : undefined;
+    const refSegFrame1 =
+        PerFrameFunctionalGroupsSequence[1] &&
```
igoroctaviano, line 654: In either case, we should avoid duplication of this ternary and move it into a new function.

src/adapters/Cornerstone/Segmentation_4X.js
igoroctaviano: It's sticky to catch, identify and create messages for all possible errors coming from dcmjs / thrid party in OHIF. Maybe dcmjs needs to define a new error layer that will throw well-defined error messages + indexes?

src/adapters/Cornerstone/Segmentation_4X.js
igoroctaviano: You could do something like getSegmentIndex(multiframe, 0) and getSegmentIndex(multiframe, 1)

igoroctaviano (changes requested):

src/adapters/Cornerstone/Segmentation_4X.js
Punzo: you are completely right!

src/adapters/Cornerstone/Segmentation_4X.js
Punzo: you are right, much cleaner and produce docs

src/adapters/Cornerstone/Segmentation_4X.js
Punzo: I agree with you @igoroctaviano , but I would do this in a separate PR.

Punzo:
@igoroctaviano comments applied a part https://github.com/dcmjs-org/dcmjs/pull/187#discussion_r589452632. I would say we can do that in a separate PR. 

P.S. for Steve :

> Generally looks good and we want the seg in idc fixed. Let me know what you think of my comments.

https://github.com/dcmjs-org/dcmjs/pull/183 from Igor already fixed the  seg in idc. This PR adds more checks and make it working the examples

examples/createSegmentationRLERoundtrip/index.html
pieper: Ah, in that case I'd probably use `labelmapBuffers` (I find it's readable to have something like `buffers.forEach( (buffer) => {} );`)

src/adapters/Cornerstone/Segmentation_4X.js
pieper: Yes, I agree.  We just did something similar for the save dialog in Slicer, where the messages from the dependency libraries are captured into message buffers that are exposed behind icons in the UI dialog.  Definitely worth its own discussion and PR.

PR #195, opened 2021-03-24
Added untilTag and includeUntilTagValue as an option for DICOM file reading

whatwilliam:
This pull request adds `untilTag` as a parameter for the `read`, `readTag`, and `readFile` functions so that the user can choose to ignore the rest of a DICOM file. A boolean flag `includeUntilTagValue` has also been added that returns the values of the given `untilTag` before ignoring the rest of the DICOM file.

PR has successfully run `npm test` and `npm run build`.

Related PR/Issues:
#134
#149 
#181

### DIFF
--- a/test/test_untilTag.js
+++ b/test/test_untilTag.js
@@ -1,8 +1,11 @@
+const expect = require("chai").expect;
 const fs = require('fs');
+const path = require("path")
 const dcmjs = require('../build/dcmjs.js')
 const buffer = fs.readFileSync('sample-dicom.dcm');
 
 const { DicomMessage, DicomMetaDictionary } = dcmjs.data;
+
 console.time('readFile');
 const fullData = DicomMessage.readFile(buffer.buffer)
 console.timeEnd('readFile');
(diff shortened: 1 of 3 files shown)


whatwilliam:
@pieper

pieper:
This looks reasonable. 👍   Would you be able to add a test?  You can reuse existing data and just confirm that you get the expected subsets of data for the different flag combinations.

whatwilliam:
The file in the test folder `test_untilTag.js` shows execution time and outputs on the data. When ignoring the untilTag by setting the `includeUntilTagValue` to `false`, having `vr` and `values` to `null` did not let the data be turned into a naturalized dataset because some operations could not be done on a `null` value, so `vr` and `values` are returned simply as `0` when the user wants to not include the untilTag.

test/test_untilTag.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs');
+const dcmjs = require('C:/Users/DHDESK/Desktop/dcmjs/dcmjs/build/dcmjs.js')
```
pieper, line 2: Obviously this hard-coded path won't work for other people 😄 

Can you integrate this with the testing system so that it will run on any machine as part of `npm test`?  Have a look at the other test files for inspiration.

test/test_untilTag.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs');
```
pieper, line 1: Have a look at how [chai] is used [in other tests](https://github.com/dcmjs-org/dcmjs/blob/master/test/test_data.js#L1) with the `expect` method so that the individual steps of the process can be verified.

test/test_untilTag.js
```diff
@@ -0,0 +1,31 @@
...
+const dcmjs = require('C:/Users/DHDESK/Desktop/dcmjs/dcmjs/build/dcmjs.js')
+const buffer = fs.readFileSync('sample-dicom.dcm');
```
pieper, line 3: Have a look at how [other tests](https://github.com/dcmjs-org/dcmjs/blob/master/test/test_data.js#L156) access data from the [data repository](https://github.com/dcmjs-org/data) to get sample data.

pieper (changes requested):
Thanks for contributing 👍 

I hope the comments are reasonable for you.  They are in the interest of making sure the code stays workable over time.  Be sure to run `npm test` with your changes in place to be sure everything else still works.

test/test_untilTag.js
whatwilliam: Whoops, didn't mean to leave that in. Should be fixed now.

test/test_untilTag.js
whatwilliam: The test has been integrated with the expect method

test/test_untilTag.js
whatwilliam: I've written the integrated test in `test_data.js` to access the arbitrary `sample-dicom.dcm` the same way as other tests that access .dcms in the test folder.

whatwilliam:
@pieper Will these changes work?

test/test_untilTag.js
pieper: Thanks for the changes, it looks good now.

Issue #194, opened 2021-03-15
should overlay data also be encapsulated like pixel data depending on transfer syntax?


codepage949:
[overlay-samples.zip](https://github.com/dcmjs-org/dcmjs/files/6139599/samples.zip)

i had two dicoms(one is 'Explicit VR Little Endian' of transfer syntax and other one is 'JPEG 2000 Image Compression (Lossless Only)'.

and i added overlay to each by dcmjs like below.

```js
let dcm = dcmjs.data.DicomMessage.readFile(...);

dcm.upsertTag(/*Overlay Rows Tag*/, 'US', ...);
dcm.upsertTag(/*Overlay Columns Tag*/, 'US', ...);
dcm.upsertTag(/*Overlay Type Tag*/, 'CS', ...);
dcm.upsertTag(/*Overlay Origin Tag*/, 'SS', [..., ...]);
dcm.upsertTag(/*Overlay Subtype Tag*/, 'LO', ...);

let overlay = dcmjs.data.BitArray.pack([...]);

if (overlay.length % 2) {
    const newOverlay = new Uint8Array(overlay.length + 1);

    newOverlay.set(overlay);
    newOverlay.set([0], overlay.length);

    overlay = newOverlay;
}

dcm.upsertTag(/*Overlay Data Tag*/, 'OB', [overlay.buffer]);
fs.writeFileSync(..., Buffer.from(dcm.write()));
```

as result, overlay was visible on 'Explicit VR Little Endian' dicom in weasis and aliza viewer but not on 'JPEG 2000 Image Compression (Lossless Only)'.

and i found that overlay data was encapsulated and had very different length from what i inserted on 'JPEG 2000 Image Compression (Lossless Only)' dicom.

is it correct behavior?

('JPEG 2000 Image Compression (Lossless Only)' dicom was not opened by dcmjs with error "Error: Request more than currently allocated buffer", so i checked it with dicom parser)

pieper:
> ('JPEG 2000 Image Compression (Lossless Only)' dicom was not opened by dcmjs with error "Error: Request more than currently allocated buffer", so i checked it with dicom parser)

It sounds like this is the core issue, right?  Does this message appear before you do the overlay part or when you try to re-read the one you created?

[...]

PR #199, opened 2021-04-16
#194 encapsulation is applied for only PixelData

codepage949:
after this fix, overlay is visible on compressed image transfer syntax.

it seems enough that encapsulation is applied for only PixelData.

is there another case to apply encapsulation?


codepage949:
from #194

pieper:
Hi - thanks for digging into this and that solution looks simple.  But I haven't got a clear idea whether it would have side effects on other studies.

Can you write a test following the format of the other tests in dcmjs?  Were you able to run the existing test suite with your change?

codepage949:
@pieper added test and passed with other tests, please review.

pieper:
Thanks for adding the test 👍 

Since this is in a deep part of the code that I know people are using in production I'm trying to be extra thoughtful about merges.  @Punzo and @igoroctaviano could you have a quick look to see if you think this is okay to merge?

pieper:
@codepage949 could you provide some extra context for this?

* what is the target use case?
* where do these overlays come from? 
* what system is expected to read them?
* how do other toolkits handle this data?  Have you tried pydicom, dcmtk, gdcm, others?

Thanks in advance

codepage949:
* what is the target use case?

    * denoting areas over image

* where do these overlays come from?

    * in my case, made it by myself. please refer to [#194](https://github.com/dcmjs-org/dcmjs/issues/194)

* what system is expected to read them?

    * ideally, any dicom viewer that complies with standards

* how do other toolkits handle this data? Have you tried pydicom, dcmtk, gdcm, others?

    * in pydicom, overlay data is inserted without any transformation even when transfer syntax is what needs encapsulation.

[side note]

i opened a dicom file which has private tag of UN vr with dcmjs and saved without any changes.

and then, opened dicom file saved by dcmjs on weasis, aliza viewer, it failed to open.

i guess it is because dcmjs apply encapsulation on private tag of UN vr too. [tag.js#L111](https://github.com/dcmjs-org/dcmjs/blob/91c70dd12e3c42f7fc542a195ab759ad763196b5/src/Tag.js#L111)

codepage949:
according to [this](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.html), i interpreted that pixel data is the only one that can be encapsulated.

pieper:
> according to [this](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.html), i interpreted that pixel data is the only one that can be encapsulated.
> 
> > Pixel Data (7FE0,0010) may be encapsulated or native.
> 
> > Overlay Data (60xx,3000) and other tags mentioned in the document.
> > ... shall be encoded in Little Endian.

Yes, I see your logic now @codepage949.

@jmhmd I think this is also related to some code you added.  Can you review and comment?

Issue #204, opened 2021-04-28
can't read an encapsulated frame whose size is greater than fragment size


codepage949:
when writing with encapsulation and fragmentMultiframe enabled, dcmjs is able to handle a frame which is greater than fragment size(20480 bytes).

but when reading, it only expect an encapsulated frame which is less than or equal to fragment size.

PR #205, opened 2021-04-28
#204 fix: can't read an encapsulated frame whose size is greater than fragment size

codepage949:


pieper:
Thanks for contributing 👍 

From what I can tell this looks like good fix.  @jmhmd looks like this touches on things you added, can you review?

Nitpick: For testing I've been trying to keep the dicom files themselves as binary release assets in https://github.com/dcmjs-org/data rather then putting them in the repo.  If you look at some of the other tests you'll find they use `downloadToFile` which is pretty easy.  I notice now that a couple of other .dcm files slipped into the testing directory but I'd like to get rid of those even though they are pretty small.

codepage949:
@pieper how to join the repo(https://github.com/dcmjs-org/data)?

PR #210, opened 2021-06-05
Fix: First pixel of a segment was omitted.

rogerbramon:
This PR proposes to fix a problem that causes the omission of the first pixel of every segment on frame. 

#### How to reproduce?
You could generate a segmentation with a square region,  when this segmentation is imported you should see a missing pixel in the top-left corner.

### DIFF
--- a/src/adapters/Cornerstone/Segmentation_4X.js
+++ b/src/adapters/Cornerstone/Segmentation_4X.js
@@ -1113,7 +1113,7 @@ function insertPixelDataPlanar(
         //
         for (let j = 0, len = alignedPixelDataI.data.length; j < len; ++j) {
             if (data[j]) {
-                for (let x = j + 1; x < len; ++x) {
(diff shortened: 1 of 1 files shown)


pieper:
@igoroctaviano it looks like this edits a line [you added here](https://github.com/dcmjs-org/dcmjs/blame/f7495ab21879b3965d83da2ef8dac920dcdc8d03/src/adapters/Cornerstone/Segmentation_4X.js#L1116).  Can you review the logic in this PR?  🙏

igoroctaviano:
@pieper all good here 👍 

Some evidence... compare the top left pixel of both images (with / without changes):

Screen Shot 2021-06-07 at 18 33 49
Screen Shot 2021-06-07 at 18 34 10

pieper:
Thanks for the confirmation @igoroctaviano 👍

PR #217, opened 2021-09-13
fix non existing named tag MRImageFrameType in MRImageNormalizer class

helghast79:
MRImageFrameType -> MRImageFrameTypeSequence (0018,9226)


pieper:
Thanks for catching this 👍   There was an earlier version of the code where we dropped the "Sequence" from tags but later decided it was better to always use the exact terms from the standard so this must have been a holdover from the old code.

PR #219, opened 2021-09-17
Make single item arrays into both array items the item value itself

wayfarer3130:
Single item sequences like Shared Functional Groups are very convenient to access as singleton items, but others like the per-item functional groups are typically accessed as a arrays, even when they have only a single item in them.  This change preserves the "arrayness" of the items even of single-item values, but also preserves the singleton items, so that it is possible to access either:
metadata.SharedFunctionalGroupsSequence[0]
and get the same value as:
metadata.SharedFunctionalGroupsSequence

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -142,8 +142,11 @@ class DicomMetaDictionary {
                 }
 
                 if (naturalDataset[naturalName].length === 1) {
-                    naturalDataset[naturalName] =
-                        naturalDataset[naturalName][0];
+                    const sqZero = naturalDataset[naturalName][0];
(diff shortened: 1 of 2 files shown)


wayfarer3130:
@pieper  - I hadn't realized that there were two branches of dcmjs.  I've updated this now against the dcmjs-org branch.

pieper:
That's a pretty cool trick 👍 

I realize (again) that the commit message didn't have the semantic versioning prefix so there won't be a new npm package.   This will get into the next release, but if you are in a hurry to use a release make a new PR with the [Commit Message Format](https://github.com/dcmjs-org/dcmjs#for-maintainers).

PR #220, opened 2021-09-27
fix: 🐛 IDC Re #2003: fix regression in parsing segmentation orietations

Punzo:
Fix regression in https://github.com/OHIF/Viewers/issues/2003

### DIFF
--- a/src/utilities/orientation/nearlyEqual.js
+++ b/src/utilities/orientation/nearlyEqual.js
@@ -24,7 +24,7 @@ export default function nearlyEqual(a, b, epsilon) {
     } else if (a == 0 || b == 0 || absA + absB < Number.EPSILON) {
         // a or b is zero or both are extremely close to it
         // relative error is less meaningful here
-        return diff < epsilon * Number.EPSILON;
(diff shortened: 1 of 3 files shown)


Punzo:
@pieper could you please metge when you have time? thanks!

fedorov:
Nice! Hopefully this will also take care of https://github.com/OHIF/Viewers/issues/2500!

src/utilities/orientation/nearlyEqual.js
```diff
@@ -24,7 +24,7 @@ export default function nearlyEqual(a, b, epsilon) {
     } else if (a == 0 || b == 0 || absA + absB < Number.EPSILON) {
         // a or b is zero or both are extremely close to it
         // relative error is less meaningful here
-        return diff < epsilon * Number.EPSILON;
+        return diff < epsilon;
```
igoroctaviano, line 27: If we have a constant (number.epsilon) and another epsilon variable, can we rename the variable to differentiate them? maybe this could help to avoid future confusion / new bugs.

PR #221, opened 2021-09-30
fix(dcmjs): Remove a circular reference to the naturalize methods

wayfarer3130:
The previous version of the code generated circular references, which some people didn't like, and also the fix didn't create an "isArray" instance because:
a = {} 
Array.isArray(Object.assign(a,[a]))==false
but:
arrayOfA = [a]
Array.isArray(Object.assign(arrayOfA,a))===true
which makes the result iterable.
@rodolfo.costa@radicalimaging.com

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -143,9 +143,14 @@ class DicomMetaDictionary {
 
                 if (naturalDataset[naturalName].length === 1) {
                     const sqZero = naturalDataset[naturalName][0];
-                    naturalDataset[naturalName] = sqZero;
-                    if (sqZero && typeof sqZero === "object") {
-                        Object.assign(sqZero, [sqZero]);
+                    if (
+                        sqZero &&
+                        typeof sqZero === "object" &&
+                        !sqZero.length
+                    ) {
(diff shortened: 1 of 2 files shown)


src/DicomMetaDictionary.js
```diff
@@ -143,9 +143,14 @@ class DicomMetaDictionary {
 
                 if (naturalDataset[naturalName].length === 1) {
```
ladeirarodolfo, line 144: @wayfarer3130 
Apart from circular reference there is a critical change to consider.
From 0.18.8 to 0.18.9 (minor release), has occurred a breaking change. I.e 
Data type for  single sequences is different to  non single sequences.

@pieper  is this expected from lib point of view? Is there a way to avoid having breaking changes released under minor releases?

pieper: > @pieper is this expected from lib point of view? Is there a way to avoid having breaking changes released under minor releases?

I hadn't realized this was a breaking change - we should reconsider this.  I'm leaning toward removing the concept of any dicom `*Sequence` being anything but an array.  It would mean more `[0]`s in the code but it would be more consistent with the standard.  Since this would break the API it should be a new minor release at least.

PR #229, opened 2021-11-02
prevent arraybuffers from being excluded of naturalized datasets

helghast79:


pieper (changes requested):
Thanks for the contribution 👍 

I think the fix makes sense, but would like @wayfarer3130 to have a look.

Two requests:
* Can you expand on the commit message to describe why this was problematic and why your patch fixes the issue.  We now have several people editing the code and we need to help each other understand what we are all doing.
* Can you change the commit message to follow [semantic-commit format](https://github.com/dcmjs-org/dcmjs#for-maintainers) so the patch release is generated automatically.  Force push the fork if you need to.

src/DicomMetaDictionary.js
```diff
@@ -149,7 +149,8 @@ class DicomMetaDictionary {
                     if (
                         sqZero &&
                         typeof sqZero === "object" &&
-                        !sqZero.length
+                        !sqZero.length &&
+                        !(sqZero instanceof ArrayBuffer)
```
pieper, line 153: @wayfarer3130 can you look and test?

helghast79:
Hi Pieper,

Sure, problem arise when loading a dicom-seg files which trigger `RangeError: Invalid typed array length`. After debugging found that the addAccessor's function is being called with the arraybuffer in question and since it doesn't have any key, it silently exists this function leaving the property with undefined value.

I believe addAccessor's function is not intended for arrays or arraybuffers so the additional filter makes sense I guess. 

Captura de ecrã 2021-11-03, às 08 07 58

src/DicomMetaDictionary.js
wayfarer3130: That makes sense for Seg, or anything else that decodes direct to raw buffers, still would like to see a unit test.

pieper:
Thanks for the context 👍   I agree a test would help.

Another small thing I think there cannot be a space between the `fix` and the `(` in the commit message header.

src/DicomMetaDictionary.js
helghast79: addAccessors function is fine, the problem comes from the conditions to which it's being called. Before, it would make `naturalDataset[naturalName] = naturalDataset[naturalName][0] `and now it leaves it as is which will breaks downstream code and make it undefined.

Have no ideia why it made sense before to remove the element out of the array but code was made with that assumption and the proposed safeguard just tries to make it backward compatible.

anyway, I'll put a test later if it helps

jsaarija:
Hi! What's the progress on this? I'm currently encountering this very issue in 0.19.6 with PixelData in CT images being placed inside an array after the changes to sequence naturalization. I currently have a workaround for this but naturally I would prefer to see this fix merged to master.

pieper:
@wayfarer3130 can you comment on this PR to see if it's good to merge?  Specifically how does it relate to #251 which seems to touch the same code.

PR #230, opened 2021-11-02
Add nrrd data to src and create corresponding test for debugging

shanebenlolo:
Hello,

I am trying to create a part 10 dicom file which can hold some .nrrd data that is parsed using ITK.js. I have been running into an issue where I set the PixelData attribute of my dataset, and receive the following error:
```
No value representation given for PixelData
```
I am not sure where to place the value representation for my pixel data, any help on the issue would be appreciated. 

**To Reproduce**
```
npm install
npm run build
npm test
```

I have commented out all other tests so they do not clutter the results. you should be met with the following output: 
```
***** Testing Nrrd *****

Truncating value 1.9531249999999996 of PixelSpacing because it is longer than 16
Truncating value 1.9531249999999996 of PixelSpacing because it is longer than 16
No value representation given for PixelData
Invalid vr type ox - using OW
Invalid vr type ox - using OW
dcmjs\build\dcmjs.js:776
            this.view = new DataView(this.buffer);
```
The file `src/nrrd/decoded_nrrd_file.json` contains arbitrary data which is in the same format one would receive if they parsed an .nrrd file using ITK. I have removed most of the pixel data to minimize the files size and make debugging a simpler process.


pieper:
Were you able to solve the issue and create the dicom instance?

shanebenlolo:
> Were you able to solve the issue and create the dicom instance?

Hey Steve,

I did not solve this exact problem, but it's because I realized what I was asking for was actually unnecessary. I thought I needed to convert an NRRD CT scan into a stack of dicom slices, but I actually ended up needing to create a labelmap3d from an NRRD segmentation file which corresponded to an existing dicom CT file.

 Once I had the NRRD file in in the form of a labelmaps3D object, it was just a matter of passing it to `dcmjs.adapters.Cornerstone.Segmentation.generateSegmentation()`

Going from the NRRD segmentation file to a labelmaps3D object required me to generate the labelmap2D buffers and metadata arrays by hand though, as `cornerstoneTool.segmentationMoudle.setters.labelmap3DByFirstImageId` was not properly generating those two fields of data from my NRRD segmentation data, so I'll probably be making a PR to cornerstoneTools about that in the near future.

pieper:
Great, I'm glad you were able to power through and find a solution 👍

shanebenlolo:
Thanks! Wouldn't have been possible without this fantastic tool! 😃

PR #233, opened 2022-01-11
Delimitation element handling

hamish-mcc:
Relevant DICOM Standard excerpt https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_7.5.html


hamish-mcc:
Requires further investigation

Issue #231, opened 2021-11-12
TypeError: Cannot redefine property: Alphabetic


sedghi:
I'm getting the following error on latest `dcmjs` version, any idea? I can confirm `dcmjs 0.18.6` does not produce this error

```js
dcmjs.es.js?3946:2743 Uncaught (in promise) TypeError: Cannot redefine property: Alphabetic
    at Function.defineProperty (<anonymous>)
    at eval (dcmjs.es.js?3946:2743)
    at Array.forEach (<anonymous>)
    at addAccessors (dcmjs.es.js?3946:2742)
    at eval (dcmjs.es.js?3946:2913)
    at Array.forEach (<anonymous>)
    at Function.naturalizeDataset (dcmjs.es.js?3946:2869)
    at Array.map (<anonymous>)
```

pieper:
Thanks for the report 👍 

Could you provide a PR with a test that replicates this issue?  

@wayfarer3130 can you have a look?

wayfarer3130:
Will look as soon as the test is available.

sedghi:
Sorry It might take me a while to create a test for this issue as the repo I'm working on is private.

sedghi:
@pieper @wayfarer3130 we debugged this today with James and Erik, and here are the results: 

Apparently you cannot `naturalizeDataset twice`, so following code will through an error. (We were naturalizing the dataset in the internals of one of our providers)

```js
const instance = DicomMetaDictionary.naturalizeDataset(instanceMetaData);
const another = DicomMetaDictionary.naturalizeDataset(instanceMetaData);
```

but the following code works 

```js
const instance = DicomMetaDictionary.naturalizeDataset(
   JSON.parse(JSON.stringify(instanceMetaData))
);
const another = DicomMetaDictionary.naturalizeDataset(
  JSON.parse(JSON.stringify(instanceMetaData))
);
```
Is there any where in the `naturalizeDataset` that it is mutating the `metadata` by reference?

pieper:
That sounds like a regression, probably introduced by [this commit](https://github.com/dcmjs-org/dcmjs/commit/70b24332783d63c9db2ed21d512d9f7b526c5222) or similar.

Calling `naturalizeDataset` should be idempotent in my opinion, especially if it was that way before.

It would be good to have a test of that along with a fix of course.  @wayfarer3130 are you able to work on that?

wayfarer3130:
[...]

PR #237, opened 2022-01-28
fix(dcmjs): #231 Fix the normalize Alphabetic twice

wayfarer3130:
Fixed issue 231 where normalizing twice fails, and prevents side affects for modifying the array of length one, but doesn't fix any of the issues modifying longer non-sequence arrays, nor modifications to the properties of the child objects.


test/test_data.js
```diff
@@ -115,6 +115,14 @@ function makeOverlayBitmap({ width, height }) {
 }
 
 const tests = {
+    test_array_items: () => {
+        const dicomJSON = JSON.stringify(require('./arrayItem.json'));
+        const datasets = JSON.parse(dicomJSON);
+        const natural0 = DicomMetaDictionary.naturalizeDataset(datasets[0]);
+        // Shouldn't throw an exception
+        const natural0b = DicomMetaDictionary.naturalizeDataset(datasets[0]);
```
pieper, line 123: Can you also test if they are the same?

pieper (changes requested):
Can you add some more description in your commit message?  Why did the previous code lead to the issue, what changes did you make and why did they fix the problem?  It will help to have more context if this or a related issue comes up again.

test/test_data.js
wayfarer3130: Yes, done.

wayfarer3130:
> Can you add some more description in your commit message? Why did the previous code lead to the issue, what changes did you make and why did they fix the problem? It will help to have more context if this or a related issue comes up again.

Redid the commit message when I added the other fix and updated to master.

pieper (approved):
Thanks for the updates 👍

PR #251, opened 2022-03-07
fix: images with color pallet lut not displaying properly

ladeirarodolfo:
Includes:
- When naturilazing dataset prevent adding accessors to TypedArray/ArrayBuffer types


wayfarer3130:
Please just use Array.isArray, drop the two extra files, and add a unit test.

PR #253, opened 2022-03-14
fix: 🐛 IDC2733: find segmentations reference source image Ids

Punzo:
Improve find logic for segmentations reference source image Ids as discussed in https://github.com/OHIF/Viewers/issues/2733.

To find the refence source image id now we do:
1) we filter out source series with the segmentation ReferencedSeriesSequence.SeriesInstanceUID
2) we filter out source instances with the segmentation FrameOfReferenceUID
3) we filter out source instances with the segmentation PerFrameFunctionalGroup.PlanePositionSequence.ImagePositionPatient (within a tolerance)

If a source instance is not found for the segmentation frame, we do a last try using simply the segmentation frame ReferencedSOPInstanceUID

### DIFF
--- a/src/adapters/Cornerstone/Segmentation_4X.js
+++ b/src/adapters/Cornerstone/Segmentation_4X.js
@@ -378,7 +378,6 @@ function generateToolState(
             throw new Error(
                 "Segmentations orthogonal to the acquisition plane of the source data are not yet supported."
             );
-            break;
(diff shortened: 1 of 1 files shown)


Punzo:
@pieper could you please review? Thanks

pieper (approved):
Makes sense - thank you Davide 👍

PR #254, opened 2022-03-20
Use the Jest testing framework and switch to GitHub Actions

richard-viney:
Thanks for the work you've done on this library!

I've recently encountered some DICOMs that have issues reading & writing in dcmjs due to what appears to be some bugs in parsing and multi frame serialisation, and plan to open a couple PRs for these with fixes. However I noticed initially that there was no test runner for the current tests, and so added [Jest](https://jestjs.io) as it's a common choice these days and makes writing additional best-practice tests more straightforward.

The main changes in this PR are:

- Add Jest as a test runner
- Migrate expect syntax to the Jest style
- Migrate tests with async behaviour to use async/await
- The actual code in the tests has been changed as little as possible
- Remove previous ad-hoc test runner
- Upgrade Babel version
- Add set of recommended VS Code extensions, including one for easily running/debugging Jest tests

Thoughts on adopting Jest as a test runner?


pieper:
Thanks for the contribution 👍 

I don't know much about testing frameworks, which is why there was just a silly stub so far, so this will be a big improvement.  I did have one suggestion for [vitest](vitest.dev) but if you have a preference for jest we can go with that.

In general it looks good - is it ready for review/merge?

richard-viney:
Vitest is fairly new and looks promising, though perhaps less relevant for a non-frontend project such as dcmjs. I actually use Vite itself (but not vitest) every day and it's great.

Jest is the most commonly used JS testing framework by some distance, and has a 93% satisfaction ranking here: https://2021.stateofjs.com/en-US/libraries/testing. Hence is a reasonable choice, and anyone wanting to contribute to dcmjs will likely be at home with it too.

If there's a desire to adopt vitest in future (its runner is a little bit nicer than Jest in some ways) then the transition would be mainly changing configs and adding some imports, because vitest supports the same describe/it/expect syntax as Jest so there'd be essentially no code changes.

Yes this PR is merge-able right now, I've reviewed and tested it all locally.

pieper:
This looks good to me, but the circleci test isn't working now and it's blocking the merge.  I could override it for now, but we'd want to have it working going forward.  @richard-viney do you have any ideas about that?

richard-viney:
I've pushed a couple of updates to the CircleCI config, but can't actually see any logs from the CircleCI runs .. are they available?

If you like I could switch it to use GitHub Actions, might be simpler having it all in one place?

pieper:
@richard-viney Yes, I would really like to see us switch to github actions and improve the testing.  Can you add that to this PR?

richard-viney:
The conversion to GitHub Actions is done. The new workflows need to be approved first before they will actually run.

The `NPM_TOKEN` needs to be setup as a GitHub secret by a repo admin. The CircleCI integration hook also needs to be removed by an admin.

I haven't been able to directly test the release workflow, but may be best to just merge and see what, if anything, breaks. There might be a permissions issue, I'm not 100% certain.

richard-viney:
I've added a set of recommended extensions, including one for the `firsttris.vscode-jest-runner` VS Code extension that lets you really easily run individual tests and set breakpoints in the IDE.

pieper:
@richard-viney thanks for your work on this.  Since it looks like you are still committing and this hasn't been reviewed yet I changed to draft for now.  Let us know when you think it's ready to go and we can go through setting up the keys and everything.  I haven't had a chance to look into any of that.

richard-viney:
This is ready to go. I've squashed it into two commits: one that adds Jest and one that switches to GitHub Actions.

To setup the secret for GitHub Actions (which would be best done prior to merging):

- Go to https://github.com/dcmjs-org/dcmjs/settings/environments
- Create an environment called `publish`
- Configure it to only be accessible to the `master` branch
- Add an `NPM_TOKEN` secret to the new `publish` environment

Please let me know if there's anything else I can do.

richard-viney:
Please let me know if I can help with the review/merge process any further here. I've got a couple of bug fixes to submit along with new-style tests for them that would be great to get the ball rolling on as well :-).

pieper:
@igoroctaviano said he can take a look at this since he has done similar work on other projects.

.gitignore
```diff
@@ -8,3 +8,6 @@ yarn-error.log
 yarn.lock
 .DS_Store
 build
+
+CTPelvis.zip
```
igoroctaviano, line 12: These seem very specific to me? can we use a temp dir instead?

test/anonymizer.test.js
```diff
@@ -0,0 +1,30 @@
...
+import fs from "fs";
+
+const { DicomMessage } = dcmjs.data;
+const { cleanTags } = dcmjs.anonymizer;
+
+it("test_export", () => {
+    expect(typeof cleanTags).toEqual("function");
+});
+
+it("test_anonymization", () => {
```
igoroctaviano, line 11: Although we wrote poor test descriptions in the past I think we should move forward with better test descriptions by actually describing the test. Adding contextual information through 'describe' blocks is a good practice too. 

@richard-viney I don't want to inherit the tech so, for now, can you please add test instructions? or code standards for testing? this could help us keep our test codebase in better shape.

@pieper Another thing I have noticed is that we don't have code coverage and I think we should have it configured + make it part of our pipeline. I'll create an issue for it later on. No worries right now. 😄

igoroctaviano: https://github.com/dcmjs-org/dcmjs/issues/259

igoroctaviano (commented):
LGTM 🎖️ 
Thank you for the great work!

I'll just update the project config to add the envs to test the actions and I'll get back to you.

test/anonymizer.test.js
pieper: 💯 for code coverage and better test descriptions.

igoroctaviano:
@richard-viney publish works (tested in a separate fork). 
Can you please resolve the conflicts before we can move forward?

test/anonymizer.test.js
richard-viney: Test instructions and best practices have been added to `README.md`.

pieper:
This looks good, but it's not clear to me how to interpret the test results.  It looks like lots of error messages but then the tests are listed as passing.

https://github.com/dcmjs-org/dcmjs/runs/6097327896?check_suite_focus=true

@igoroctaviano should we go ahead and merge and then remove the circleci from the required checks.

richard-viney:
AFAICT the `console.error()` calls haven't changed and happen identically on current master. However, Jest takes the extra step/precaution of printing a full stack trace for a `console.error()` call, which does blow up the test logs a bit. The tests themselves could definitely be tidied up somewhat, but I think that's outside the scope of this PR which is for switching to Jest and GitHub Actions.

One of my followup PRs is a fix for the `Invalid vr type ...` error that is caused by problems in the `BinaryRepresentation` class.

pieper:
Thanks for the explanation 👍   I'll go ahead and merge and we can sort out what to do next.

PR #261, opened 2022-04-27
feat: Add Cornerstone3D adapter for Length tool

sedghi:
- Adds the initial work for a Cornerstone3D adapter 
- Fixes the prettierrc which was missing

@JamesAPetts Can you review please


JamesAPetts (commented):
Compared with old implementation, LGTM.

PR #264, opened 2022-05-04
WIP: feat: Bidirectional Arrow and EllipticalROI adapters for CS3D

sedghi:
@JamesAPetts can you please also review this as well?

### DIFF
--- a/src/adapters/Cornerstone3D/ArrowAnnotate.js
+++ b/src/adapters/Cornerstone3D/ArrowAnnotate.js
@@ -1,0 +1,111 @@
+import MeasurementReport from "./MeasurementReport.js";
+import TID300Point from "../../utilities/TID300/Point.js";
+import CORNERSTONE_3D_TAG from "./cornerstone3DTag";
+
+const ARROW_ANNOTATE = "ArrowAnnotate";
+const CORNERSTONEFREETEXT = "CORNERSTONEFREETEXT";
+const trackingIdentifierTextValue = "Cornerstone3DTools@^0.1.0:ArrowAnnotate";
+
+class ArrowAnnotate {
+    constructor() {}
+
+    static getMeasurementData(MeasurementGroup, imageId, imageToWorldCoords) {
+        const {
+            defaultState,
+            SCOORDGroup,
+            findingGroup
+        } = MeasurementReport.getSetupMeasurementData(MeasurementGroup);
+
+        const text = findingGroup.ConceptCodeSequence.CodeMeaning;
+
+        const { GraphicData } = SCOORDGroup;
+
+        const worldCoords = [];
+        for (let i = 0; i < GraphicData.length; i += 2) {
+            const point = imageToWorldCoords(imageId, [
+                GraphicData[i],
+                GraphicData[i + 1]
+            ]);
+            worldCoords.push(point);
+        }
+
+        const state = {
+            ...defaultState,
+            toolType: ArrowAnnotate.toolType,
+            data: {
+                text,
+                handles: {
+                    points: [worldCoords[0], worldCoords[1]],
+                    activeHandleIndex: 0,
+                    textBox: {
+                        hasMoved: false
+                    }
+                }
+            }
+        };
+
+        return state;
+    }
+
+    static getTID300RepresentationArguments(tool, worldToImageCoords) {
+        const { data, metadata } = tool;
+        let { finding, findingSites } = tool;
+        const { referencedImageId } = metadata;
+
+        if (!referencedImageId) {
+            throw new Error(
+                "ArrowAnnotate.getTID300RepresentationArguments: referencedImageId is not defined"
+            );
+        }
+
+        const { points } = data.handles;
+
+        const pointsImage = points.map(point => {
+            const pointImage = worldToImageCoords(referencedImageId, point);
+            return {
+                x: pointImage[0],
+                y: pointImage[1]
+            };
+        });
+
+        const TID300RepresentationArguments = {
+            points: pointsImage,
+            trackingIdentifierTextValue,
+            findingSites: findingSites || []
+        };
+
+        // If freetext finding isn't present, add it from the tool text.
+        if (!finding || finding.CodeValue !== CORNERSTONEFREETEXT) {
+            finding = {
+                CodeValue: CORNERSTONEFREETEXT,
+                CodingSchemeDesignator: "CS3DT1",
+                CodeMeaning: data.text
+            };
+        }
+
+        TID300RepresentationArguments.finding = finding;
+
+        return TID300RepresentationArguments;
+    }
+}
+
+ArrowAnnotate.toolType = ARROW_ANNOTATE;
+ArrowAnnotate.utilityToolType = ARROW_ANNOTATE;
+ArrowAnnotate.TID300Representation = TID300Point;
+ArrowAnnotate.isValidCornerstoneTrackingIdentifier = TrackingIdentifier => {
+    if (!TrackingIdentifier.includes(":")) {
+        return false;
+    }
+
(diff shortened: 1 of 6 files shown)


src/adapters/Cornerstone3D/ArrowAnnotate.js
```diff
@@ -0,0 +1,111 @@
...
+            points: pointsImage,
+            trackingIdentifierTextValue,
+            findingSites: findingSites || []
+        };
+
+        // If freetext finding isn't present, add it from the tool text.
+        if (!finding || finding.CodeValue !== CORNERSTONEFREETEXT) {
+            finding = {
+                CodeValue: CORNERSTONEFREETEXT,
+                CodingSchemeDesignator: "CS3DT1",
```
JamesAPetts, line 81: Can we enumerate and write a comment about this code somewhere, otherwise its alien to anyone.

src/adapters/Cornerstone3D/Bidirectional.js
```diff
@@ -0,0 +1,203 @@
...
+                Math.pow(firstPointPairs[0][2] - firstPointPairs[1][2], 2)
+        );
+
+        let secondPointPairsDistance = Math.sqrt(
+            Math.pow(secondPointPairs[0][0] - secondPointPairs[1][0], 2) +
+                Math.pow(secondPointPairs[0][1] - secondPointPairs[1][1], 2) +
+                Math.pow(secondPointPairs[0][2] - secondPointPairs[1][2], 2)
+        );
+
+        let lengthPointPairs;
```
JamesAPetts, line 127: I know these were called this before in the old implementation, but can we call these `longAxis` and `shortAxis` instead? I think its semantically better and easier to follow that they become the long and short axes.

src/adapters/Cornerstone3D/EllipticalROI.js
```diff
@@ -0,0 +1,167 @@
...
+const ELLIPTICALROI = "EllipticalROI";
+const FINDING = "121071";
+const FINDING_SITE = "G-C0E3";
+
+const trackingIdentifierTextValue = "Cornerstone3DTools@^0.1.0:EllipticalROI";
+
+class EllipticalROI {
+    constructor() {}
+
+    // TODO: this function is required for all Cornerstone Tool Adapters, since it is called by MeasurementReport.
```
JamesAPetts, line 14: Shall we remove this TODO? Or at least write that a base class with an abstract method would be cleaner?

src/adapters/Cornerstone3D/EllipticalROI.js
```diff
@@ -0,0 +1,167 @@
...
+        const {
+            defaultState,
+            NUMGroup,
+            SCOORDGroup
+        } = MeasurementReport.getSetupMeasurementData(MeasurementGroup);
+
+        const { GraphicData } = SCOORDGroup;
+
+        // GraphicData is ordered as [majorAxisStartX, majorAxisStartY, majorAxisEndX, majorAxisEndY, minorAxisStartX, minorAxisStartY, minorAxisEndX, minorAxisEndY]
+        // But Cornerstone3D points are ordered as top, bottom, left, right for the
```
JamesAPetts, line 25: > so we need to identify if the majorAxis is horizontal or vertical

Do we? Won't CS3D just auto correct this when you drag a point, and all stat calculations should be correct anyway iirc?

JamesAPetts (approved):
Some minor comments but LGTM.

src/adapters/Cornerstone3D/EllipticalROI.js
sedghi: done

src/adapters/Cornerstone3D/EllipticalROI.js
sedghi: No they are always top, bottom, left, right when I checked

src/adapters/Cornerstone3D/Bidirectional.js
sedghi: sure

sedghi:
@pieper ready for review :)

sedghi:
Need to fix a bug in ellipse for oblique

src/adapters/Cornerstone3D/EllipticalROI.js
```diff
@@ -24,60 +30,74 @@ class EllipticalROI {
...
+        if (!imagePlaneModule) {
+            throw new Error("imageId does not have imagePlaneModule metadata");
+        }
+
+        const { columnCosines } = imagePlaneModule;
+
+        // find which axis is parallel to the columnCosines
+        const columnCosinesVec = vec3.fromValues(...columnCosines);
+
+        const projectedMajorAxisOnColVec = vec3.dot(
```
JamesAPetts, line 66: I think it would be safer to normalize these vectors: `majorAxisVec` and `minorAxisVec`.

Later we just compare them to EPSILON, which is not a fair comparison if they aren't unit vectors. If the ellipse is really big, the comparison to ~0 could break.

src/adapters/Cornerstone3D/EllipticalROI.js
```diff
@@ -24,60 +30,74 @@ class EllipticalROI {
...
+            majorAxisVec
+        );
+
+        const projectedMinorAxisOnColVec = vec3.dot(
+            columnCosinesVec,
+            minorAxisVec
+        );
+
+        let ellipsePoints = [];
+        if (projectedMajorAxisOnColVec < EPSILON) {
```
JamesAPetts, line 77: Don't we have to take the absolute value of the dot product here? What if the major axis is along the direction cosine, but it points in the opposite direction?

JamesAPetts: Also this if statement is saying "If the major axis _doesn't_ project on the column vector, insert it along the row (left&right coordinates of ellipse)", followed by another if that says "If the minor axis _doesn't_ project on the column vector, insert it along the row (left&right coordinates of ellipse)".

These negative statements are a bit confusing, its generally an antipattern to do "if not this, else this", when it can be better read as "if this, else this" by swapping the statements around.

It would be better to check if "if the major axis _does_ project on the column vector, insert it along the column vector (top&bottom coordinates of ellipse)", which can be done by the comparison:

```
// Normalize the majorAxisVec earlier
vec3.normalize(majorAxisVec, majorAxisVec)

// This can now only ever be a number from -1 to 1
const projectedMajorAxisOnColVec = vec3.dot(
      columnCosinesVec,
      majorAxisVec
);

// Take absolute.
const absoluteOfMajorDotProduct = Math.abs(projectedMajorAxisOnColVec)


if (Math.abs(absoluteOfMajorDotProduct - 1) < EPSILON) {
        // ... major axis vec is along column cosines vec...
}
```

JamesAPetts (changes requested):
There are some holes in the ellipse calculation that need fixing due to using non normalized vectors.

Otherwise some minor usability comments.

src/adapters/Cornerstone3D/index.js
```diff
@@ -3,13 +3,16 @@ import Length from "./Length.js";
...
+import { CORNERSTONEFREETEXT, CodingSchemeDesignator } from "./CodingScheme";
 
 const Cornerstone3D = {
     Length,
     Bidirectional,
     EllipticalROI,
     ArrowAnnotate,
-    MeasurementReport
+    MeasurementReport,
+    CORNERSTONEFREETEXT,
```
JamesAPetts, line 14: And then following the other comment, export the `CodingScheme` here.

JamesAPetts (approved):
Minor export comment otherwise LGTM.

JamesAPetts (approved):
LGTM.

PR #266, opened 2022-05-06
Fix the circle adapter to prevent writing undefined

wayfarer3130:
The circle adapter was writing undefined for the area and diameter.

### DIFF
--- a/src/adapters/Cornerstone/CircleRoi.js
+++ b/src/adapters/Cornerstone/CircleRoi.js
@@ -29,8 +29,9 @@ class CircleRoi {
             active: false,
             cachedStats: {
                 area: NUMGroup.MeasuredValueSequence.NumericValue,
-                radius: 5,
-                perimeter: 3
+                // Dummy values to be updated by cornerstone
+                radius: 0,
+                perimeter: 0
             },
             handles: {
                 end: {
--- a/src/adapters/Cornerstone/CircleRoi.js
+++ b/src/adapters/Cornerstone/CircleRoi.js
@@ -17,19 +17,21 @@ class CircleRoi {
             NUMGroup,
             SCOORDGroup
         } = MeasurementReport.getSetupMeasurementData(MeasurementGroup);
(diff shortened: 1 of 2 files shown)


pieper (approved):
I have no problem with this.  Does anyone from the cornerstone group wish to review?  @sedghi @JamesAPetts

src/adapters/Cornerstone/CircleRoi.js
```diff
@@ -29,7 +28,9 @@ class CircleRoi {
             toolType: CircleRoi.toolType,
             active: false,
             cachedStats: {
-                area: NUMGroup.MeasuredValueSequence.NumericValue
+                area: NUMGroup.MeasuredValueSequence.NumericValue,
+                radius: 5,
```
wayfarer3130, line 32: Needed dummy values here - eventually should use the right set of values read from the response.

src/adapters/Cornerstone/CircleRoi.js
```diff
@@ -67,8 +68,9 @@ class CircleRoi {
     static getTID300RepresentationArguments(tool) {
         const { cachedStats, handles, finding, findingSites } = tool;
         const { start: center, end } = handles;
-        const { area } = cachedStats;
+        const { area, radius } = cachedStats;
```
wayfarer3130, line 71: Read the values as specified.

sedghi:
> @sedghi - could you respond to Steve please about this issue? It is just an adapter that works for cornerstone 2d circle roi, as the previous one had some bugs in it.

I would ask @JamesAPetts since he knows better this space, my only comment is that why we are hard coding radius: 5

src/adapters/Cornerstone/CircleRoi.js
wayfarer3130: Changed to be 0 @sedghi  - added a comment to the code explaining why it has the value 0 as it is jsut a dummy.

JamesAPetts:
LGTM. I agree with setting the values 0 if we don't know them, its the only honest thing to do.

Issue #263, opened 2022-04-28
naturalizeDataset in v0.18.11 creates in memory objects of 300mb+ 


dbousamra:
Hi all.

I upgraded dcmjs, and I noticed calls to naturalizeDataset were taking longer. Looking further into it, writing the response to disk gives files of 300mb+, whereas in 0.18.10 (and prior), they were ~ 4kb. The file is full of thousands of `null`'s.

pieper:
That's interesting.  Can you reproduce this with any data you can share?

dbousamra:
Here is a code sandbox with a simple reproduction: https://codesandbox.io/s/recursing-booth-61xnc5?file=/src/index.ts

If you notice, the output is: `JSON lengths { rawTags: 5149, naturalizedTags: 300153806 }`. The `naturalizedTags` are being filled with many null values (presumably tags that aren't in my file, but dcmjs is eagerly appending).

If you switch the version back to pre 0.18.11, you will see output is: `JSON lengths { rawTags: 5149, naturalizedTags: 3734 }`. The null values aren't present. 

I hope this helps you debug the issue. Please let me know if you need anything else.

pieper:
Thanks for the code sandbox but I keep getting this 502 bad gateway error.  Does this also happen for you?  Is there another way to test the code?

![image](https://user-images.githubusercontent.com/126077/166113212-7ce59ec0-0894-44b7-ba27-1236e5e8e079.png)

pieper:
If we have a simple node script that can reproduce and detect the error someone could try git bisect on it.

I'm not sure what change would lead to this issue, but I'm wonder if it's an unintended consequence of the accessors.  @wayfarer3130 could you have a look?

https://github.com/dcmjs-org/dcmjs/compare/v0.18.1...v0.21.0#diff-a70070702108a92f492970cff5b5ef645d5c33c56c01a775d65063e9bdbefa7aR157

dbousamra:
@pieper try just Ctrl-S/Cmd-S in the editor. It will force a save, and you should see the script run

dbousamra:
I believe the issue was introduced in this changeset: https://github.com/dcmjs-org/dcmjs/compare/v0.18.10...v0.18.11

pieper:
Thanks for narrowing this down @dbousamra, yes, that's the commit I was thinking might at the heart of it.

[...]

PR #267, opened 2022-05-06
Fix/263 large objects #263

wayfarer3130:
Fix large objects being created by the addAccessor.  The problem was that an accessor was added for a private creator tag, which ends up being an integer, so it extended the array to the value of the private tag.  This caused a huge number of nulls to be created.  The solution was to use a proxy rather than accessors, to avoid actually having the values be native accessors, but done as a request time call.  This also reduces the memory footprint overall.


wayfarer3130:
@pieper - I modified the approach.  I think it is a better approach overall, as it is a lower memory footprint.
Added some extra tests to cover a few more things.

pieper:
> @pieper - I modified the approach. I think it is a better approach overall, as it is a lower memory footprint.
> Added some extra tests to cover a few more things.

Great, thanks @wayfarer3130 👍 

@dbousamra can check if this fixes #263 for you?

dbousamra:
@pieper Not easily no. I will need to wait for a release. If that sample test case I provided passes, I am sure it will be fine

pieper:
@wayfarer3130 are you able to test with @dbousamra 's example data?  Alternatively, are you sure enough of the fix that we should just do a release?

wayfarer3130:
@pieper I added @dbousamra test data as a unit test, after deleting the patient name details.  It was failing before the fix and passing afterwards, so I'm reasonably confident this fixes things.  I also added a couple of other tests to try to get more coverage on this area as it is clearly something that is tricky to get right.

pieper:
Thanks @wayfarer3130.  @dbousamra please test when you can.

PR #268, opened 2022-05-18
Feat/cs3d planar freehand sr

JamesAPetts:
In this PR I do multiple things:
- Overhaul the interface of the Cornerstone3D `MeasurementReport` adapter so that it is more natural and doesn't rely on extra knowledge of the system. The old interface you had to have some knowledge of the order of the annotations in the SR (i.e. you had to parse the SR before... parsing the SR) as there was an array of imageIds that are in an assumed order. Now you just pass in a map of sopInstanceUID to imageId.
- Allow for annotations to be from multiple referenced series in one SR. The correct references are made so that we reference each series correctly, rather than having all measurements incorrectly reference the first series of the first measurement (this is only done for the Cornerstone3D adapter, but effort has been put in to ensure that the old Cornerstone adapter still works as it did).
- Add read/write for Cornerstone3D annotations: planar freehand tool and probe tool.

### DIFF
--- a/src/adapters/Cornerstone3D/MeasurementReport.js
+++ b/src/adapters/Cornerstone3D/MeasurementReport.js
@@ -93,6 +93,57 @@ export default class MeasurementReport {
         }
     }
 
+    static generateDatasetMeta() {
+        // TODO: what is the correct metaheader
+        // http://dicom.nema.org/medical/Dicom/current/output/chtml/part10/chapter_7.html
+        // TODO: move meta creation to happen in derivations.js
+        const fileMetaInformationVersionArray = new Uint8Array(2);
+        fileMetaInformationVersionArray[1] = 1;
+
+        const _meta = {
+            FileMetaInformationVersion: {
+                Value: [fileMetaInformationVersionArray.buffer],
+                vr: "OB"
+            },
+            //MediaStorageSOPClassUID
+            //MediaStorageSOPInstanceUID: sopCommonModule.sopInstanceUID,
+            TransferSyntaxUID: {
+                Value: ["1.2.840.10008.1.2.1"],
+                vr: "UI"
+            },
+            ImplementationClassUID: {
+                Value: [DicomMetaDictionary.uid()], // TODO: could be git hash or other valid id
+                vr: "UI"
+            },
+            ImplementationVersionName: {
+                Value: ["dcmjs"],
+                vr: "SH"
+            }
+        };
+
+        return _meta;
+    }
+
+    static generateDerivationSourceDataset(
+        StudyInstanceUID,
+        SeriesInstanceUID
+    ) {
+        const _vrMap = {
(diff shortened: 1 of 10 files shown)


src/adapters/Cornerstone3D/Probe.js
```diff
@@ -0,0 +1,109 @@
...
+class Probe {
+    constructor() {}
+
+    static getMeasurementData(
+        MeasurementGroup,
+        sopInstanceUIDToImageIdMap,
+        imageToWorldCoords,
+        metadata
+    ) {
+        console.log("TODO");
```
sedghi, line 17: console log

src/adapters/Cornerstone3D/MeasurementReport.js
```diff
@@ -77,7 +77,79 @@ function getMeasurementGroup(
 export default class MeasurementReport {
     constructor() {}
 
-    static getSetupMeasurementData(MeasurementGroup) {
+    static getCornerstoneLabelFromDefaultState(defaultState) {
+        const { findingSites = [], finding } = defaultState;
+
+        let freeTextLabel = findingSites.find(
+            fs => fs.CodeValue === "CORNERSTONEFREETEXT"
```
sedghi, line 84: can we use CodeScheme from here (src/adapters/Cornerstone3D/CodingScheme.js) for the `CORNERSTONEFREETEXT`

src/adapters/Cornerstone3D/MeasurementReport.js
```diff
@@ -100,31 +172,43 @@ export default class MeasurementReport {
...
-            frameIndex: ReferencedFrameNumber || 1,
-            complete: true,
-            finding: findingGroup
-                ? findingGroup.ConceptCodeSequence
-                : undefined,
-            findingSites: findingSiteGroups.map(fsg => {
-                return { ...fsg.ConceptCodeSequence };
-            })
+            annotation: {
+                annotationUID: DicomMetaDictionary.uid(),
```
sedghi, line 191: shouldn't this be the same as annotationUID in cs3d? (I don't konw, just asking)

sedghi (approved):
Great improvements, see my minor comments

src/adapters/Cornerstone3D/MeasurementReport.js
JamesAPetts: good catch

src/adapters/Cornerstone3D/MeasurementReport.js
JamesAPetts: The annotationUIDs in CS3D are for internal referencing, this just generates one from the SR's data. We aren't currently encoding the annotationUID on save right now, its something we could do in the future. Maybe its useful, but i'm not sure. Happy to discuss at some point.

PR #270, opened 2022-05-23
fix: ArrowAnnotateTool adapter in Cornerstone3D parsing label

sedghi:
- Expose constants for Cornerstone3D 
- Fix tool state generation for Arrow annotate
- Fix the Backward compatibility with legacy cornerstone for arrow

### DIFF
--- a/src/adapters/Cornerstone3D/ArrowAnnotate.js
+++ b/src/adapters/Cornerstone3D/ArrowAnnotate.js
@@ -43,10 +43,8 @@ class ArrowAnnotate {
             worldCoords.push(point);
         }
 
-        // Since we use this adapter for both Cornerstone3D and Cornerstone legacy ArrowAnnotate,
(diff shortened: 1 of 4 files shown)


JamesAPetts (changes requested):
I didn't realise we actually encoded both points for the arrow for Cornerstone3D, we shouldn't be doing this at all.

After long conversations with David Clunie on [this document](https://docs.google.com/document/d/1bR6m7foTCzofoZKeIRN5YreBrkjgMcBfNA7r9wXEGR4/edit#bookmark=id.4pf2giqt6jd4) we concluded the measurement is actually just the point, and the rest of the stuff like the arrow's direction, should be something like presentation state.

In CSL we only encoded the point the arrow is pointing at, and we should actually do the same here. The POINT SCOORD type only expects a single point.

JamesAPetts (approved):
LGTM

JamesAPetts (approved):

PR #274, opened 2022-06-08
fix(dcmjs):Make the conceptCodeSequence generateReport a proxy

### LINKED ISSUE #273: Inconsistent data from SR generate report and naturalizing the dataset
With https://github.com/dcmjs-org/dcmjs/commit/70b24332783d63c9db2ed21d512d9f7b526c5222 we have switched to adding a [proxy](https://github.com/dcmjs-org/dcmjs/commit/70b24332783d63c9db2ed21d512d9f7b526c5222#diff-a70070702108a92f492970cff5b5ef645d5c33c56c01a775d65063e9bdbefa7aR148-R157) for objects in the naturalizing process which will make arrays of 1 items to stay as an array (but with backward compatibility of `.` notation to access keys through a Proxy)

SR generate report (`MeasurementReport.generateReport`) returns a naturalized data set which doesn't go through the process of assigning a Proxy, so we end up with objects, as opposed to array of objects (which would be the case if we just naturalizeDataSet itself)

I talked to @wayfarer3130 and he kindly accepted to investigate this.

wayfarer3130:
The conceptCodeSequence generated by the generateReport is supposed
to be consistent with that generated by parsing, but because of the
proxy it wasn't consistent, so this change makes it consistent

@sedghi - could you please check this branch out and see if it fixes your problem?
FIxes #273


src/adapters/Cornerstone/Bidirectional.js
```diff
@@ -133,10 +134,10 @@ class Bidirectional {
             toolName: "Bidirectional",
             visible: true,
             finding: findingGroup
-                ? findingGroup.ConceptCodeSequence
+                ? addAccessors(findingGroup.ConceptCodeSequence)
```
sedghi, line 137: Can't you use the same way we get the `defaultState` in the other tools?

src/utilities/addAccessors.js
```diff
@@ -29,7 +39,14 @@ const handler = {
...
 const addAccessors = (dest, sqZero) => {
-    const ret = [sqZero];
+    if (dest.__isProxy) return dest;
+    let itemZero = sqZero;
+    if (itemZero === undefined) {
+        if (typeof dest !== "object") return dest;
+        if (Array.isArray(dest) && dest.length !== 1) return dest;
+        itemZero = Array.isArray(dest) ? dest[0] : dest;
+    }
+    const ret = [itemZero];
```
sedghi, line 49: I guess these changes give `addAccessors` to do a proxy on the `object` too? Do we really need that for objects? I thought the whole point was to be able to use both `seq[0].b` and `seq.b`

sedghi (commented):
I tested it and it works, I'm just not sure why are adding a proxy to objects too.

src/utilities/addAccessors.js
wayfarer3130: The point is to make it for sq[0].b and seq.b  The point of the code is to make it easy to create a proxy on various types of objects, without needing to really know what they are - so it works on any of:
{object}
[{object}]
[]
See the unit tests I added.

wayfarer3130: It doesn't add a proxy to the object, it creates a new object that is an array and is a proxy to the passed in object.
That is,
obj = {...attr: value}
addAccessor(obj)
returns a proxy to an array that happens to have access methods for all attributes attr in obj.
That way, the returned object can be accessed either as the array or as the object itself, so that both types work.

src/adapters/Cornerstone/Bidirectional.js
wayfarer3130: The default state doesn't seem to always include the finding/findingSite data - probably I should revert just this single change, but otherwise leave setting the finding /site data to another issue.

src/utilities/addAccessors.js
sedghi: So does that mean we will end up with all fields as a Proxy? even if they are simple objects?

wayfarer3130: No, only things that are simple objects contained within a sequence need to be proxies.

sedghi (approved):
Look good to me

pieper:
@wayfarer3130 can you resolve the conflicts?

wayfarer3130:
Will resolve them after I get over covid - too out of it to fix anything
right now.

Bill

pieper:
> after I get over covid

Oof -  get well soon!

wayfarer3130:
@sedghi - I made one more change you requested - it still isn't used universally, but I'd rather have several small issues than one that fixes mroe things than needed.  Let me know if it still works for you.
Also updated the PR with the latest merges.

sedghi:
@wayfarer3130 This is great, yes it solves my issue

PR #276, opened 2022-06-09
Fix errors parsing binary tags

richard-viney:
Have run into a few bugs in the binary tag parsing on some DICOMs, and this PR contains fixes and test cases for those issues.

The two included DCM files are 13MB and 3MB - should these be hosted on GitHub releases like the other data is?

Details:

- Fix reading of OB tags with undefined length due to not handling trailing padding
- Fix reading of OB tags that use an offset table when there are large tags prior to the tag being read
- Make reading of sequence tags more explicit and add error checking
- Add two test cases for the above with DICOMs that previously had errors in parsing

### DIFF
--- a/src/BufferStream.js
+++ b/src/BufferStream.js
@@ -315,12 +315,10 @@ class BufferStream {
             throw new Error("Request more than currently allocated buffer");
         }
 
-        const newBuf = new ReadBufferStream(
-            this.buffer,
-            null,
-            this.offset,
-            this.offset + length
-        );
+        const newBuf = new ReadBufferStream(this.buffer, null, {
+            start: this.offset,
+            stop: this.offset + length
+        });
         this.increment(length);
 
         return newBuf;
@@ -341,10 +339,17 @@ class BufferStream {
 }
 
 class ReadBufferStream extends BufferStream {
-    constructor(buffer, littleEndian, start, end) {
+    constructor(
+        buffer,
+        littleEndian,
+        options = {
+            start: null,
+            stop: null
+        }
+    ) {
         super(buffer, littleEndian);
-        this.offset = start || 0;
-        this.size = end || this.buffer.byteLength;
+        this.offset = options.start || 0;
+        this.size = options.stop || this.buffer.byteLength;
         this.startOffset = this.offset;
         this.endOffset = this.size;
     }
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -365,18 +365,30 @@ class BinaryRepresentation extends ValueRepresentation {
                     offsets = [];
                 }
 
-                const getFrameOrFragement = stream => {
-                    var nextTag = Tag.readTag(stream);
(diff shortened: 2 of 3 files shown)


Punzo:
tested over OHIF and Slim on a couple of studies, no regression found


image
image

richard-viney:
Thanks, I've switched to downloading the files in the tests. I think the URLs are right but can't confirm until the data files are available in the other repo. The two data files can be downloaded here:

https://drive.google.com/file/d/1tQFRfY9PBtOeDPuAB1ZCb9TNLlWuV8Zx/view?usp=sharing

richard-viney:
URLs updated and tests pass. I think there's a typo though, "stessors" should be "stressors".

Ouwen:
@richard-viney we seem to be modifying the same files, shall I include your changes on top of #279 or vice versa?

richard-viney:
@Ouwen I think it's up to the maintainers, though you're right we have fixed at least one of the same bugs to do with incorrect stream offsets! I'm keen to get this PR in because it fixes failures parsing valid DICOM files, and adds more tests to that end. Do the tests added in this PR pass when run on your branch?

richard-viney:
After looking at rebasing this after the merge of #279 a few things came up.

Please see my comment there for details: https://github.com/dcmjs-org/dcmjs/pull/279#issuecomment-1159866553

richard-viney:
I've updated the PR and put the additional binary file here: https://drive.google.com/file/d/1DXM5ChyzPwf1co3gTChbk0QcP55FfqBc/view?usp=sharing

Ouwen:
@richard-viney thanks for going through the trouble of rebasing. The tests pass for me once the binary file is placed into my test folder.

I made a minor commit which puts the start/stop as an kwarg instead of arg. This way future options like Uint8 view only can live easy without cluttering the public API.

https://github.com/HeartLab/dcmjs/pull/2/commits/aea9c6fb312558d008b414a848b10395e51143f4

richard-viney:
@Ouwen Thanks I've put that change onto this PR, and also done some other refactoring in `BinaryRepresentation.readBytes()` to simplify the code a bit further.

Tests are all passing (once the binary-tag.dcm file is in the data repo)

I think we're OK to merge then?

Ouwen:
@richard-viney if possible could we keep the `.map` for frame/fragment parsing? The reason is that there are some advantages to treating frame parsing as a parallel map operation vs explicit for-loop.

For these new changes, would you kindly separate out the new commit changes on top of your previous PR commit for easier review? Referencing the previous commit history here: https://github.com/ouwen/dcmjs/tree/heart

richard-viney:
Sure thing, I'll submit those further changes on a separate PR and we can discuss there. Have updated this PR accordingly. Thanks.

pieper:
> URLs updated and tests pass. I think there's a typo though, "stessors" should be "stressors".

Thanks, I fixed the typo, they are now here: https://github.com/dcmjs-org/data/releases/tag/binary-parsing-stessors

PR #277, opened 2022-06-09
Allow backslashes in UIDs in order to support C-FIND in dcmjs-dimse

richard-viney:
When using this library with C-FIND requests via `dcmjs-dimse`, it is necessary to allow backslashes in UIDs because the spec says it is possible to query a list of UIDs by using a backslash as a delimiter.

Without this change UID list queries aren't possible in dcmjs-dimse because the backslashes get stripped out.

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -1019,10 +1019,26 @@ class UniqueIdentifier extends StringRepresentation {
     }
 
     readBytes(stream, length) {
-        return this.readNullPaddedString(stream, length).replace(
-            /[^0-9.]/g,
-            ""
-        );
+        const result = this.readNullPaddedString(stream, length);
+
+        const BACKSLASH = String.fromCharCode(0x5c);
+        const uidRegExp = /[^0-9.]/g;
+
+        // Treat backslashes as a delimiter for multiple UIDs, in which case an
+        // array of UIDs is returned. This is used by DICOM Q&R to support
+        // querying and matching multiple items on a UID field in a single
+        // query. For more details see:
+        //
+        // https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.2.2.2.2.html
+        // https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.4.html
+
+        if (result.indexOf(BACKSLASH) === -1) {
+            return result.replace(uidRegExp, "");
+        } else {
+            return result
+                .split(BACKSLASH)
+                .map(uid => uid.replace(uidRegExp, ""));
+        }
     }
 }
 
--- a/test/data.test.js
+++ b/test/data.test.js
@@ -8,6 +8,7 @@ import os from "os";
 import path from "path";
 import unzipper from "unzipper";
 import followRedirects from "follow-redirects";
+import { promisify } from "util";
 
 const { https } = followRedirects;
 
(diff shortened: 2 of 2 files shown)


pieper:
I haven't looked at DIMSE in a long time, but I would think that you would have a `VM` of greater than one if you have multiple UIDs in a query, and that we should keep the `UI` `VR` strictly matching the [standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html#:~:text=See%20Note%202-,UI,Unique%20Identifier%20(UID),-A%20character%20string).  That is, I think the backslash would be parsed at a different level.

richard-viney:
Yep the backslash separator isn't part of the DICOM standard itself, but dcmjs already supports it on other VR types, e.g. `IS` (integer string) and `DS` (decimal string), which is quite handy, so UID seems to be the odd one out here in not supporting it (there may also be others, I haven't done an exhaustive search).

pieper:
I didn't write this part of the code originally, so I'm reverse engineering here too, but at least for the decimal string (`DS`) case it appears that rather than treating the backslash as part of the token it splits the string and returns a list of numbers.  Wouldn't you want the same behavior for `UID`?

https://github.com/dcmjs-org/dcmjs/blob/37082a6966a4a5763cdd2cf6635a6456923dc19b/src/ValueRepresentation.js#L529-L543

richard-viney:
The splitting on backslash is already being done here:

https://github.com/dcmjs-org/dcmjs/blob/37082a6966a4a5763cdd2cf6635a6456923dc19b/src/DicomMessage.js#L226

So simply letting the backslash through the `UniqueIdentifier` parsing like this PR does is sufficient. For the decimal and integer string cases it's different because they need to convert to an actual number, or array of numbers, so the splitting is done in the `ValueRepresentation` subclass.

pieper:
I'm still not convinced we should treat a string with backslashes in it as a valid UID VR.  Can you provide more information and a test or example use case in dicom-dimse where this is required?

richard-viney:
Ok, I've compared to pydicom and it works in a similar way, i.e. it treats backslashes as UID delimiters in any DICOM content it parses. This isn't strictly correct, because it should check against the VM (value multiplicity) of the tag it's parsing. For a standard DICOM file a UID field could have a multiplicity of 1, and for DIMSE it would allow multiple. Doing this additional check would be a much larger change to dcmjs, and the only benefit is preventing multiplicity in those contexts where it isn't technically meant to be allowed. pydicom doesn't seem to have bothered with this, and just allows multiplicity anywhere, which is the approach dcmjs has also taken on DS and IS VRs, it just wasn't implemented for UID VRs.

The relevant part of the specs are:

https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.2.2.2.2.html
https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.4.html

I've updated the PR with these links, and a test for reading a DICOM that has multiple values for a UID and a DS tag.

pieper (changes requested):
This looks much better to me, thanks for taking the time to sort through it.

I realize the dcm file is very small, but for consistency could you remove it from the PR so we can put it as a release asset instead?

richard-viney:
Sure thing, the binary file has been removed and put here https://drive.google.com/file/d/180YkJ6cqHKm0RphxMGCeUR0aFsUNDL1u/view?usp=sharing

PR #278, opened 2022-06-09
Convert empty DecimalString and NumberString to null instead of zero

richard-viney:
This is needed when doing C-FIND with dcmjs-dimse, where "" means 'match everything', not zero.

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -534,15 +534,19 @@ class DecimalString extends StringRepresentation {
         if (ds.indexOf(BACKSLASH) !== -1) {
             // handle decimal string with multiplicity
             const dsArray = ds.split(BACKSLASH);
-            ds = dsArray.map(ds => Number(ds));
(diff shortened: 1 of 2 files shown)


richard-viney:
Nice catch, I've pushed a fix for it writing back "null" in the serialisation case. It now writes back the empty string again so the read=>write roundtrip with a "" value in IntegerString/DecimalString now keeps it unchanged, as one would expect (this wasn't the case before this PR, i.e it would be transformed to zero).

richard-viney:
Also worth noting that this is a breaking API change and so would need a v0.23 release.

pieper:
> Also worth noting that this is a breaking API change and so would need a v0.23 release.

Note that we use these [commit message conventions](https://github.com/dcmjs-org/dcmjs#for-maintainers) so this would trigger a new major version, which seems a bit much for this.  Maybe this is better seen as a fix, since as you pointed out the behavior was basically incorrect before.

Issue #282, opened 2022-06-24
Bug: Fragment merging results in zero value arraybuffer


Ouwen:
Recent merge doesn't type set as `Uint8Array` on ArrayBuffer when using `.set` function

PR #283, opened 2022-06-24
Refactor and fragment bug

Ouwen:
Fix for #282.

Also refactors options into dictionary params to maintain public API.
Moved optional parameter tests into separate data-options.test.js
Removed legacy untilTag.test.js


pieper:
Does this change break the API?

Ouwen:
It shouldn’t break the API (DicomMessage.readFile)

Ouwen:
@pieper the refactor commit will change the function signature of

`DicomMessage.read` and `DicomMessage.readTag`

However, my understanding is that these are largely used internal to the dcmjs package. I've done a sweep to edit any place they are used (tests).

The change puts options like `ignoreErrors`, `untilTag`, and `includeUntilTagValue`, into an `options` param. Future PRs which add options can be made without modifying the function signature.

pieper:
Yes, I agree, they are unlikely to be used outside of the parsing code.  I wonder if the "correct" thing to do would be to change the names of these and similar internal methods to start with an underscore and leave the existing ones as-is but mark them as deprecated.  What do you think?

Ouwen:
I think that makes sense and would be a good notification for anyone still using it that it'll be dropped in the next major version release. I can put up a quick PR for that.

Ouwen:
@pieper let me know if this looks good to you

pieper:
Thanks for this 👍  keeping the API stable makes me more comfortable.

PR #288, opened 2022-07-07
move "eslint-config-prettier" to devDependencies

florianlink:
It is wrong to list "eslint-config-prettier" as runtime "dependency", it needs to be a "devDependency".
Having it as runtime dependencies introduces a large amount of licenses that the license checker will detect, which are not actually part of the deployed runtime of dcmjs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MIT",
       "dependencies": {
         "@babel/runtime-corejs2": "^7.17.8",
-        "eslint-config-prettier": "^8.5.0",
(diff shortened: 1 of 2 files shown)


swederik:
Thanks! Can you run npm install so the package-lock file is updated as well?

florianlink:
Yes, I added the lock file as well. By the way, a lot of dev dependencies report that they are outdated/deprecated, probably it would make sense to upgrade to newer versions of rollup etc.

florianlink:
Can you merge this, please?

PR #291, opened 2022-07-09
added support for specific character set

florianlink:
refactored VR classes
refactored reading/writing ascii and encoded strings
changed hex writing to direct uint8 writing
reimplemented null padded read
removed unused code
unified test data loading and cache the test data locally for faster test runs

this fixes: https://github.com/dcmjs-org/dcmjs/issues/84

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "example": "examples"
   },
   "scripts": {
-    "test": "jest",
+    "test": "jest --testTimeout 60000",
     "build": "rollup -c",
     "build:examples": "npm run build && npx cpx 'build/**/*.{js,map}' examples/js",
     "start": "rollup -c -w",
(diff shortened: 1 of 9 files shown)


florianlink:
Looks like some tests have been added after my fork, I will merge them into my branch and port them as well.
By the way, all these  log.error("Invalid vr type " + type + " - using OW"); etc. make it hard to read the real errors.
Maybe we could tackle that issue as well? Many of the example datasets cause these error logs (already before my changes).

pieper:
Yes, thank you @florianlink  👍   I agree the `log.error` messages add a lot of noise.  The testing framework has been evolving lately and any suggested fixes would be very welcome.

florianlink:
I fixed the failing test, I overlooked one test case and it worked because of the previously downloaded file on my machine.

PR #294, opened 2022-07-17
fix(fragment): write padding to even length on final fragments of encapsulated frame data

### LINKED ISSUE #293: Encapsulated pixel data of odd-length writes padding byte in incorrect location
A padding byte for odd-length encapsulated binary data is currently being written at the end of writing the OB tag, but padding bytes need to be written for _each frame_, not the whole tag. That is, the last fragment for each frame should get a padding byte if that frame's data is of uneven length. This is the behaviour of pydicom.

Will open a PR containing a fix and test shortly.

richard-viney:
A padding byte for odd-length encapsulated binary data was being written at the end of writing the OB tag, but padding bytes need to be written for each frame, not the whole tag. That is, the last fragment for each frame should get a padding byte if that frame's data is of uneven length. This is also the behaviour of pydicom.

Closes #293.

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -226,7 +226,8 @@ class BinaryRepresentation extends ValueRepresentation {
             // Calculate a total length for storing binary stream
             var bufferLength = 0;
             for (i = 0; i < frames; i++) {
-                bufferLength += value[i].byteLength;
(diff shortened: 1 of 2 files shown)


pieper:
Looks very nice - thank you 🙏

PR #295, opened 2022-07-21
fix(imports): missing import for addAccessors

sedghi:
For some reason (probably a force push) the imports for https://github.com/dcmjs-org/dcmjs/pull/274 was missing. I'm pretty sure I checked the PR and either was working for me or I fixed and pushed or not pushed. This is a fix for the missing import


sedghi:
@pieper I'm merging this as it was breaking the whole library

pieper:
Thanks @sedghi 👍

PR #298, opened 2022-07-30
fix: support a charset of "ISO_IR 6", and throw errors for unsupported encodings

richard-viney:
This PR makes two changes:

- Make `ISO_IR 6` value for `SpecificCharacterEncoding` not `console.warn()` about an unsupported encoding.
- If the `SpecificCharacterEncoding` in a DICOM isn't supported by Node.js'`TextDecoder` then the error is now propagated and reading the DICOM will fail. The reason is that swallowing errors from unsupported encodings and proceeding seems like a dangerous/undesirable thing to do.


richard-viney:
Upon closer inspection there were some bigger problems with character set handling which are addressed in #299. Am closing this PR in favour of that one.

PR #299, opened 2022-07-30
fix: several issues with character set handling

richard-viney:
- Fix several character sets not working, e.g. "ISO_IR 13" and "ISO_IR 166"
- Fix warning being shown when the "ISO_IR 6" character set is used
- Add tests for reading encoded string data in several different character sets
- Throw exception on an unsupported character set or multiple character sets rather than logging and continuing on (this can be overridden by setting ignoreErrors)

The new `encodingMapping` was based on code in `pydicom` as well as the DICOM spec.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -10,8 +10,40 @@ const EXPLICIT_BIG_ENDIAN = "1.2.840.10008.1.2.2";
 const singleVRs = ["SQ", "OF", "OW", "OB", "UN", "LT"];
 
 const encodingMapping = {
+    "": "iso-8859-1",
+    "iso-ir-6": "iso-8859-1",
+    "iso-ir-13": "shift-jis",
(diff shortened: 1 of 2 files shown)


pieper:
This is looking good - thanks for the contribution 👍 

Let me know when it's ready to merge.

richard-viney:
The only additional change I think is warranted is to throw an exception on an attempt to use multiple character sets, because this isn't currently supported.

If ignoreErrors is set to true we would allow parsing to proceed and emit a warning similar to the existing one.

Seem reasonable?

richard-viney:
I've added the changes to throw an exception on multiple character sets as well.

Setting `ignoreErrors` to true will allow the code to proceed past any character set issue.

richard-viney:
Good to merge now 👍

pieper (approved):
Thanks for the contribution 👍

PR #301, opened 2022-08-08
fix exceptions caused by undefined cached stats in adapters. Safe pro…

ladeirarodolfo:
…gramming fix only

### Motivation
CachedStats property of annotation tools is populated async. So, to prevent JS exceptions when adapting report we should add some safe programming conditions.

###Includes
- Fix JS exception for adapting Tools (related to cachedStats)
- Safe programming only (no feature added)

fixes https://github.com/dcmjs-org/dcmjs/issues/302


pieper (approved):
LGTM 👍

PR #303, opened 2022-08-10
feat(anonymizer) - export Array tagNamesToEmpty and modify cleanTags

juliencastelneau:
- export Array tagNamesToEmpty => Modify (add or remove) values before calling cleanTags
- add dict as parameter to function cleanTags => Choose which tags you want to modify for de-identification 
For instance, a call to anonymizer should be:
```js
               const { cleanTags, tagNamesToEmpty } = dcmjs.anonymizer;

                var tagsToReplace = {
                    "00100010":   "ANON^PATIENT",
                    "00100020":   "ANON^ID"
                }
                var seriesDescription = "SeriesDescription";
                if (tagNamesToEmpty.indexOf(seriesDescription) != -1) {
                    tagNamesToEmpty.splice(tagNamesToEmpty.indexOf(seriesDescription), 1);
                } 
                var studyDescription = "StudyDescription";
                if (tagNamesToEmpty.indexOf(studyDescription) == -1) {
                    tagNamesToEmpty.push(studyDescription);
                } 

                cleanTags(dicomDict.dict, tagsToReplace);
```

### DIFF
--- a/src/anonymizer.js
+++ b/src/anonymizer.js
@@ -2,7 +2,7 @@ import log from "./log.js";
 import { DicomMetaDictionary } from "./DicomMetaDictionary.js";
 import { Tag } from "./Tag.js";
 
-var tagNamesToEmpty = [
+export var tagNamesToEmpty = [
     // please override these in specificReplaceDefaults to have useful values
     "PatientID",
     "PatientName",
@@ -235,19 +235,16 @@ var tagNamesToEmpty = [
     "DataSetTrailingPadding"
 ];
 
-export function cleanTags(dict) {
+export function cleanTags(dict, tagNamesToReplace) {
     tagNamesToEmpty.forEach(function (tag) {
         var tagInfo = DicomMetaDictionary.nameMap[tag];
         if (tagInfo && tagInfo.version != "PrivateTag") {
             var tagNumber = tagInfo.tag,
                 tagString = Tag.fromPString(tagNumber).toCleanString();
             if (dict[tagString]) {
-                log.log("empty tag " + tag);
                 var newValue;
-                if (tagString == "00100010") {
-                    newValue = ["ANON^PATIENT"];
-                } else if (tagString == "00100020") {
-                    newValue = ["ANONID"];
+                if (tagString in tagNamesToReplace) {
+                    newValue = [tagNamesToReplace[tagString]];
                 } else {
                     newValue = [];
                 }
--- a/src/index.js
+++ b/src/index.js
@@ -41,7 +41,7 @@ import adapters from "./adapters/index.js";
 import utilities from "./utilities/index.js";
 import sr from "./sr/index.js";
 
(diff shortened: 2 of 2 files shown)


src/anonymizer.js
```diff
@@ -2,7 +2,7 @@ import log from "./log.js";
 import { DicomMetaDictionary } from "./DicomMetaDictionary.js";
 import { Tag } from "./Tag.js";
 
-var tagNamesToEmpty = [
+export var tagNamesToEmpty = [
```
pieper, line 5: I'm thinking there could be side effects here - one piece of code modifies the anonymization feature and then later it might be used by another piece of code that expects it to be in the default state.

Maybe instead we a a method to return a copy of the list and then pass in the lists as optional arguments to `cleanTags` (with the internal defaults being used if not specified).

pieper (changes requested):
Thanks for working on this 👍

src/anonymizer.js
juliencastelneau: ok thx. I will work on it soon

juliencastelneau: done

juliencastelneau:
Hello, 
Is there an issue with the PR ?
@pieper  I made the changes you requested. 
Maybe I do not understand something. Do you want another modification ?
thx

pieper:
Thanks for the contribution and the reminder 👍

pieper:
@juliencastelneau it looks like there's at least one error due to the anonymizer - could you have a look?

https://github.com/dcmjs-org/dcmjs/actions/runs/3047634658/jobs/4911834982#step:5:2486

juliencastelneau:
The test fails because I added an argument to the method cleanTags
I think there are 2 ways to solve.

1. Modify the test:

Add the lines below in anonymizer.test.js (line 26)
```
    var tagsToReplace = {
        "00100010":   "ANON^PATIENT",
    }
    
    // when
    cleanTags(dicomDict.dict, tagsToReplace);

```
instead of 
```
    cleanTags(dicomDict.dict);
```
2. Modify the method cleanTags to ensure the behaviour is unchanged:

```
export function cleanTags(
    dict,
    tagNamesToReplace = undefined,
    customTagNamesToEmpty = undefined
) {
    
    if (tagNamesToReplace==undefined) {
         tagNamesToReplace = {
            "00100010":   "ANON^PATIENT",
            "00100020":   "ANON^ID"
         }
    }
    
    var tags =
        customTagNamesToEmpty != undefined
            ? customTagNamesToEmpty
            : tagNamesToEmpty;
``` 
@pieper The best way is to modify the method cleanTags I guess.
Do I need to make a new PR ?

pieper:
Thanks for looking into this.  I think the second option is better since it maintains backwards compatibility for the API.  We should also add a new test demonstrating the ways to use the method and confirm it keeps working.

Yes, a new PR would be great since this has already been merged.

PR #304, opened 2022-08-26
fix: 🐛 relax condition in nearlyEquals check for detecting numbers near to zero

Punzo:
Fix https://github.com/OHIF/Viewers/issues/2869, see https://github.com/OHIF/Viewers/issues/2869#issuecomment-1228331103 for more info

### DIFF
--- a/src/utilities/orientation/nearlyEqual.js
+++ b/src/utilities/orientation/nearlyEqual.js
@@ -18,10 +18,10 @@ export default function nearlyEqual(a, b, epsilon) {
     const absA = Math.abs(a);
     const absB = Math.abs(b);
     const diff = Math.abs(a - b);
-    if (a == b) {
+    if (a === b) {
         // shortcut, handles infinities
         return true;
-    } else if (a == 0 || b == 0 || absA + absB < Number.EPSILON) {
+    } else if (a === 0 || b === 0 || absA + absB < epsilon * epsilon) {
         // a or b is zero or both are extremely close to it
         // relative error is less meaningful here
         return diff < epsilon;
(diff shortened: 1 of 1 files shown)


Punzo:
@pieper can you please review and merge, please?

pieper:
@Punzo thanks for quick action on this 👍 

Let's review it together on Thursday to make sure we fully understand what we want the behavior to be here.  At the least we should update the comment since we are no longer following the logic of the source function from here: https://floating-point-gui.de/errors/comparison/.  The current comment implies the code is still just a transliteration to javascript.

What I want to get a handle on is what a valid [value of tolerance](https://github.com/dcmjs-org/dcmjs/blob/c209047861d2ef6c17526c6d1e29d5db7e895b29/src/adapters/Cornerstone/Segmentation_4X.js#L279) should be for data we reasonably expect to be displayed.

Since a common use case seems to be going to and from 32-bit float in a nifti1 header I'd like to take into account what rounding errors we are likely to see and set the tolerance to something that will be reasonable and robust.

Punzo:
> @Punzo thanks for quick action on this +1
> 
> Let's review it together on Thursday to make sure we fully understand what we want the behavior to be here. At the least we should update the comment since we are no longer following the logic of the source function from here: https://floating-point-gui.de/errors/comparison/. The current comment implies the code is still just a transliteration to javascript.
> 
> What I want to get a handle on is what a valid [value of tolerance](https://github.com/dcmjs-org/dcmjs/blob/c209047861d2ef6c17526c6d1e29d5db7e895b29/src/adapters/Cornerstone/Segmentation_4X.js#L279) should be for data we reasonably expect to be displayed.
> 
> Since a common use case seems to be going to and from 32-bit float in a nifti1 header I'd like to take into account what rounding errors we are likely to see and set the tolerance to something that will be reasonable and robust.

Sure, let's discuss it next Thursday at the OHIF-IDC call.

pieper (approved):

PR #307, opened 2022-09-12
Fix unit of sr num content item

hackermd:
The value of the `MeasurementUnitsCodeSequence` attribute got incorrectly included into `NumContentItem`.


hackermd:
Thanks @pieper for the fast turnaround.

PR #308, opened 2022-09-14
[FIX & TESTS] cleanTags : check if param is undefined. Add 3 test

juliencastelneau:
This PR fixes a test failure due to PR #303 
1. Check if a param is passed in function cleanTags. If not, create it and fill with default value.
It maintains backwards compatibility for the API.

2. Add 3 tests demonstrating the ways to use the method 

- force value for a specific tag. Usefull for pseudo-anonymization
- force a tag to be kept  => Remove it in the default array `tagNamesToEmpty`
- force a tag to be removed => Add it in the default array `tagNamesToEmpty`

### DIFF
--- a/src/anonymizer.js
+++ b/src/anonymizer.js
@@ -241,9 +241,15 @@ export function getTagsNameToEmpty() {
 
 export function cleanTags(
     dict,
-    tagNamesToReplace,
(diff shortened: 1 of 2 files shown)


pieper (approved):
Thank you 👍

PR #312, opened 2022-09-17
feat(deflated): Added support for reading datasets with deflated transfer syntax

PantelisGeorgiadis:


pieper (changes requested):
Looks very good and thanks for adding the test 👍 

My only request is that you remove the test data from here and instead we attach it as an asset to a release in [the data repo](https://github.com/dcmjs-org/data) so we can keep this repository small.

PantelisGeorgiadis:
Thank you for your review @pieper! 
You can find the deflated sample file I used [here](https://www.dclunie.com/images/compressed/deflate_tests_release.tar.gz).
I would be grateful if you push the zip file to the data repo so I can update this PR.

pieper:
Great - here's the release with the files as zip.

https://github.com/dcmjs-org/data/releases/tag/deflate-transfer-syntax

PantelisGeorgiadis:
Thank you @pieper! I updated the tests to pull the deflated sample files from the data repo.

pieper (approved):
Thanks!

PR #317, opened 2022-10-03
Resolves #316 Cornerstone3D adaptor - Multiframe support - add "frameNumber" to the Annotation.data

### LINKED ISSUE #316: Support annotations on multi-frame files through Cornerstone3D
Annotations made by Cornerstone3D tools are being parsed by the adapter for Cornerstone3D.

And we want to be able to add annotations to specific frames of multi-frame files.
Now in the annotation data parsed by the adapter, we don't have any data that can related to the frame number to which the annotation is attached.

The frame number information is needed in the annotation data.

md-prog:
Cornerston3D's Annotation data type allows extending "data" field (key:value).

We are going to use the field to add "frameNumber" information to annotations.
This will help supporting multi-frame DICOM files on Cornerstone.

### DIFF
--- a/src/adapters/Cornerstone3D/ArrowAnnotate.js
+++ b/src/adapters/Cornerstone3D/ArrowAnnotate.js
@@ -78,7 +78,7 @@ class ArrowAnnotate {
                     hasMoved: false
                 }
             },
-            frameNumber: ReferencedFrameNumber || 1,
+            frameNumber: ReferencedFrameNumber
         };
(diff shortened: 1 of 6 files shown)


pieper:
Adding cross-ref: Fixes #316

pieper:
Makes sense to me.  It would be good if someone from the CS team reviewed.  Perhaps you could have a look @sedghi ?

PR #321, opened 2022-10-09
Create Node version constraint at >= v11

### LINKED ISSUE #319: Crash due to no TextEncoder on nodejs
Following https://github.com/dcmjs-org/dcmjs/commit/bfd6693e7491b17a2e615393eb5b80af3a78de9b nodejs now crashes with undefined use of [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder).  According to [this post](https://stackoverflow.com/questions/69187442/const-utf8encoder-new-textencoder-in-node-js) it is fixed with the node util package like this:
```
const util = require('util');
this.encoder = new util.TextEncoder("utf-8");
```
And indeed this works when I hack it into the build.  It needs to be integrated so that the code will work on both node and browser out of the box.

vjau:
Should fix #319 
This change put a hard constraint on Node version number : npm should prevent users from installing if node version is under `v11` which is the first version that has `TextEncoder` in the global object like in the browser.
Changes in `package.json` set the node version constraint in the `engines` entry.
The `engine-strict` directive in the added `.npmrc` file should tell npm to enforce the constraint at npm install time.
I modified documentation accordingly.
Alternatively, we could just limit the constraint to a warning in the documentation and let users install at their own risk.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ _Parts of DICOM that dcmjs *will not* focus on:_
 
 ## In Node
 
+We aim to support at least the `Maintenance` version according to the [Node.js release schedule](https://github.com/nodejs/release#release-schedule).
+
+At the moment, it should work with Node 11 and above (although it's not recommended to use End Of Life versions).
+
 ```None
 // To install latest _stable_ release
 npm install --save dcmjs
--- a/package.json
+++ b/package.json
@@ -2,6 +2,10 @@
   "name": "dcmjs",
   "version": "0.24.0",
   "description": "Javascript implementation of DICOM manipulation",
+  "engines" : {
+    "node" : ">=11.0.0"
+  },
+
   "main": "build/dcmjs.js",
   "module": "build/dcmjs.es.js",
   "directories": {


pieper (approved):
Thanks 👍

pieper:
The netlify checks are failing now, saying the node version is too old.  

E.g. from https://app.netlify.com/sites/dcmjs/deploys/634281592ae0f90008b99d5c :
```
4:08:09 AM: Installing NPM modules using NPM version 6.13.4
4:08:10 AM: npm ERR! code ENOTSUP
4:08:10 AM: npm ERR! notsup Unsupported engine for dcmjs@0.24.0: wanted: {"node":">=11.0.0"} (current: {"node":"8.17.0","npm":"6.13.4"})
4:08:10 AM: npm ERR! notsup Not compatible with your version of node/npm: dcmjs@0.24.0
4:08:10 AM: npm ERR! notsup Not compatible with your version of node/npm: dcmjs@0.24.0
4:08:10 AM: npm ERR! notsup Required: {"node":">=11.0.0"}
4:08:10 AM: npm ERR! notsup Actual:   {"npm":"6.13.4","node":"8.17.0"}
4:08:10 AM: npm ERR! A complete log of this run can be found in:
4:08:10 AM: npm ERR!     /opt/buildhome/.npm/_logs/2022-10-09T08_08_10_287Z-debug.log
4:08:10 AM: Error during NPM install
4:08:10 AM: Build was terminated: Build script returned non-zero exit code: 1
```
Looks like @dannyrb removed the pinned version but perhaps we need to add it back now: https://github.com/dcmjs-org/dcmjs/commit/3e2978be96fc111d920018d4afe5f9bd5b3e355d

Issue #306, opened 2022-09-12
[Cornerstone adapters] Tool is undefined when saving FreehandRoi tool in structured report


SveinungOverland:
https://github.com/dcmjs-org/dcmjs/blob/78528325e16624e86e205c7bf6a334e18dc279e1/src/adapters/Cornerstone/FreehandRoi.js#L46-L47

Hey, we use DCMjs to create structured reports from cornerstone and cornerstone-tools (not the new cornerstone3d).
For the most part this seems to work just fine, but we recently wanted to add FreehandRoi tooling to the reports. This never seemed to work as tool was always undefined when saving the report. Upon further examination the code seems to work just fine if `tool` is put into the argument list.
The question is really just, why is `tool` commented out? The function seems to work just fine with it in, and not work at all without it.

pieper:
Thanks for the report @SveinungOverland.

@wayfarer3130 can you comment?

https://github.com/dcmjs-org/dcmjs/commit/c65b93000a31b72b2a3ef3aede03838463dd4264#diff-e912a5f9f4a457613ce5a4cfb6d57552cb1bf3b38f8f07a56ea8501ca41bfea5

wayfarer3130:
@SveinungOverland  - I'm not sure why it was commented out - it shouldn't have been, but for that class I was just updating it to match the format of the other ones, without having a full configuration that enabled it to run.  Please create a PR with it added back end, testing both save and restore.

PR #322, opened 2022-10-10
Removed comment around getTID300RepresentationArguments 'tool' parameter

SveinungOverland:
Fix for issue #306. Missing tests but since dcmjs seems to be missing a testing setup for adapters I think someone more familiar with js testing and dcmjs should set the standard.


pieper:
Thanks for PR.  I agree more testing should happen, but realistically testing this requires cornerstone so for this reason and others it would make sense for these adapters to migrate out of the dcmjs core someday.

PR #323, opened 2022-10-11
Export TID300 - Point class

md-prog:
This is to prepare dcmjs library to separate Cornerstone Adapter as a separate package.

https://github.com/dcmjs-org/dcmjs/pull/320#issuecomment-1275178945

In Cornerstone3D adapter code, the `utilities/TID300/Point` is being used, but this one is not publicly exported from dcmjs.


wayfarer3130:
It would be nice to get this in, as it allows separating out the cornerstone adapter piece, which should reduce chatter between the two projects as only core fixes will need to be included after that.

pieper (approved):
LGTM

PR #331, opened 2022-12-13
fix: infinite loop on dcm with no meta length

### LINKED ISSUE #326: Reading malformed DICOM file causes infinite loop
Loading the following malformed DICOM file results in an infinite loop in the dcmjs parser:

https://drive.google.com/file/d/1LySSmhpfNPLHAYR9WObIHL_NlpkX9a-i/view?usp=sharing

This happens because the meta length is assumed to be present and correctly formed, but in this DICOM file it is absent. Interestingly, pydicom loads the file without any errors. I haven't investigated exactly how/why it is achieving this.

conrad-scherb:
Fixes #326 and contains a test to check that reading is abandoned if no meta length is present.


pieper:
Thanks for the fix and the test 👍

PR #335, opened 2023-01-12
fix:Add more utilities - compression, addAccessor, orientation

wayfarer3130:
There are some additional classes needed to externalize the adapters.  No functional change.


wayfarer3130:
@pieper - could we get this merged so I can proceed with moving the adapters?  This is the first part

PR #337, opened 2023-01-13
fix(semantic): Indicate that the review name needs the semantic

wayfarer3130:
Arg - @pieper - still wasn't quite right, it looks like it tests the first line of the overall commit message to master, NOT the individual commits, and that requires that the PR review name have the right message.  Hopefully this will be the last one!!!


pieper:
A can completely empathize - the merge history for dcmjs has several of these semantic version issues that bite me from time to time.  But it is kind of handy when it works.

PR #346, opened 2023-04-17
Fix: reading sequences with undefined length items - length value of item delimitation tag and new item tag will be used to detect the length of the sequence more precise

kursawero:
The problem:

A DICOM Data Element Value item with undefined length is closed by an Item Delimitation Tag at the end. The previous implementation (ValueRepresentation.js) uses just the tags (4 byte) to for a stack counter that increases by one with every new Item Tag and decreases by one with every Delimitation Tag. In some of our data sets, we have binary data that by coincidence have one or more similar byte sequences. In that case, the end of a sequences is not detected and the parsing fails, because the while loop runs until the end of the data stream.

The fix:

The nema standard defines fixed undefined length attributes after the new item tag and a four byte zero length attribute after the item delimitation tag. We are reading the length attribute in both cases and increasing/decreasing the stack only if the length attribute value is the same as in the standard. Comparing 8 byte instead of 4 byte is more precise and it is more unlikely that binary data have similar sequences by coincidence.
 
Nema standard definition:

https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_7.5.2.html

I added two tests as propsals. 

One is a unit test which is a little bit limited, because the readBytes method does a lot. If you do not like the constances, we can put the values directly into the DicomDataReadBufferStreamBuilder. I had in mind to reuse the constants in the code, but it seems a bit tricky, because you would need to cast them for using it in if statements for instance.

The other is a kind of integration test with a specific file that is already part of your reporitory. The "it" folder would be the place for integration tests. I'm open if you prefer another folder structure, but I think it would be good to separate the different tests a little bit.


src/constants/dicom/Tags.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * DICOM tags as constants
+ */
+module.exports = Object.freeze({
+    SEQUENCE_DELIMITATION_TAG: 0xfffee0dd,
+    ITEM_TAG: 0xfffee000,
+    ITEM_DELIMITATION_TAG: 0xfffee00d
```
pieper, line 7: These are already in dictionary.js so I don't think we should duplicate them here.

src/constants/dicom/uids/TransferSyntaxUID.js
```diff
@@ -0,0 +1,6 @@
+module.exports = Object.freeze({
+    IMPLICIT_LITTLE_ENDIAN: "1.2.840.10008.1.2",
+    EXPLICIT_LITTLE_ENDIAN: "1.2.840.10008.1.2.1",
+    DEFLATED_EXPLICIT_LITTLE_ENDIAN: "1.2.840.10008.1.2.1.99",
+    EXPLICIT_BIG_ENDIAN: "1.2.840.10008.1.2.2"
```
pieper, line 5: These are unfortunately defined multiple places in the code already.  It does make sense to centralize them and use them consistently everywhere. 

Another stylistic thing, I think it would be easier to have a smaller number of files, so maybe just one dicomConstants.js for anything that's not proper to the data dictionary.

test/helper/DicomDataReadBufferStreamBuilder.js
```diff
@@ -0,0 +1,65 @@
...
+        while (hexValueItemArray.length < 4) {
+            hexValueItemArray = ['0x00'].concat(hexValueItemArray);
+        }
+
+        this.itemArray = this.itemArray.concat(hexValueItemArray);
+    }
+
+    addUlExampleItem() {
+        const fileMetaInfoGroupLengthTag = ['0x02', '0x00', '0x00', '0x00'];
+        // const fileMetaInfoGroupLengthVr = 'UL';
```
pieper, line 48: Dead code to be removed?

test/helper/DicomDataReadBufferStreamBuilder.js
```diff
@@ -0,0 +1,65 @@
...
+        const length = ['0x04', '0x00', '0x00', '0x00'];
+        const item = ['0x04', '0x00', '0x00', '0x00'];
+
+        this.itemArray = this.itemArray.concat(fileMetaInfoGroupLengthTag);
+        // this.itemArray.push(fileMetaInfoGroupLengthVr.charCodeAt(0).toString(16), fileMetaInfoGroupLengthVr.charCodeAt(1).toString(16));
```
pieper, line 53: Dead code?

test/it/DicomMessage.readFile.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+    describe('test some tags from the meta section', () => {
+        const metaTags = dicomDict['meta'];
+
+        it('has the TransferSyntaxUID', () => {
+            expect(metaTags).toHaveProperty('00020010.Value[0]', '1.2.840.10008.1.2');
+        })
+
+    })
+})
```
pieper, line 39: Thanks so much for the test 👍 

I like the idea of breaking up the tests a bit to keep things organized.  The folder name `it` isn't clear to me though.  I'd prefer calling it `integration` or `integrationTests`.

pieper (changes requested):
Thank you for digging in a fixing this 👍 

I haven't worked with any data like this, but I hope it's not common since, as you point out, there's still a chance that the data would include the delimitation markers by chance.  But since this is part of the standard it's good that we try to get things right.

I'm guessing you've checked these files with other toolkits (dcmtk, pydicom for example) to verify these fixes match what they do?

test/it/DicomMessage.readFile.test.js
kursawero: I think the naming is really up to the team that organizes the code. At least, I did not found a common design pattern yet. 

The "it" was used in some articles, like here: [Split unit and integration tests](https://blog.worldline.tech/2020/04/10/split-unit-and-integration-tests.html). 

I like "integration" as well. I changed the folder from "it" to "integration".

test/helper/DicomDataReadBufferStreamBuilder.js
kursawero: Sorry, I overlooked it. - it is removed now.

test/helper/DicomDataReadBufferStreamBuilder.js
kursawero: Sorry, I overlooked it. - it is removed now.

src/constants/dicom/Tags.js
kursawero: The dictionary would not work for this specific use case, because the idea was to have the name as constant in the code instead of the hex value or in case of the dictionary the tag as "(FFFE,E000)" String, for instance. I moved the 3 items into the class as constants, since we are not using them somewhere else. I hope that is ok for you.

src/constants/dicom/uids/TransferSyntaxUID.js
kursawero: I moved such items to one file ./src/constants/dicom.js. 

Having it in a separate folder is a bit personal preference of me, but in our own projects, we make more use of constants and I have more dedicated files. Would keep such decision up to you. I can also move it to the ./src folder as dicomConstants.js if you like that more.

kursawero:
Thank you very much for the quick review. I tried to address your comments. Please feel free to resolve the items or if you want changes please reply.

I added the changes as fixups, means we need to squash the commits before it can be merged. I hope that helps to follow the changes easier and we can revert changes easier if necessary.

src/constants/dicom/uids/TransferSyntaxUID.js
pieper: Yes, I like this form just fine.  Makes more sense to me.

test/it/DicomMessage.readFile.test.js
pieper: Yes, this is great.

src/constants/dicom/Tags.js
pieper: I'm not understanding why the dictionary wouldn't be used -  these three the tags are in the dictionary, e.g. https://github.com/dcmjs-org/dcmjs/blob/master/src/dictionary.js#L30767. It's not a big deal, but it would be nice to have things localized.

If you need the hex value it would be nice to add a method to the DicomMetaDictionary for that (like the [punctuate and unpunctuate](https://github.com/dcmjs-org/dcmjs/blob/master/src/DicomMetaDictionary.js#L14-L28) ones).

kursawero: I added methods and tests to the DicomMetaDictionary that we can use it to get the tags by name. Now, we get the tags from the dictionary.js. I hope that was, what you had in mind.

pieper: Yes, thanks!

PR #348, opened 2023-04-20
refactor(): use TransferSyntaxUID constants from /constants/dicom.js …

kursawero:
…in DicomMessage.js, Tag.js, data.test.js

The /constants/dicom.js was introduced recently ([link to the pull request](https://github.com/dcmjs-org/dcmjs/pull/346)). 

This pull request introduces a refactoring where some classes use the constants from this file to have the definition in a central place. In addition, the linter fixed some small formating issues automatically. I hope the linter settings are correct for this project.

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -1,14 +1,15 @@
-import { ReadBufferStream } from "./BufferStream.js";
-import { DeflatedReadBufferStream } from "./BufferStream.js";
-import { Tag } from "./Tag.js";
-import { DicomMetaDictionary } from "./DicomMetaDictionary.js";
+import { DeflatedReadBufferStream, ReadBufferStream } from "./BufferStream.js";
+import {
+    DEFLATED_EXPLICIT_LITTLE_ENDIAN,
+    EXPLICIT_BIG_ENDIAN,
+    EXPLICIT_LITTLE_ENDIAN,
+    IMPLICIT_LITTLE_ENDIAN
+} from "./constants/dicom";
 import { DicomDict } from "./DicomDict.js";
+import { DicomMetaDictionary } from "./DicomMetaDictionary.js";
+import { Tag } from "./Tag.js";
 import { ValueRepresentation } from "./ValueRepresentation.js";
 
-const IMPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2";
-const EXPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2.1";
-const DEFLATED_EXPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2.1.99";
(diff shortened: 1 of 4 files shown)


pieper:
I like the idea, but it looks like there are now circular dependencies and some other issues.

https://app.netlify.com/sites/dcmjs/deploys/6440d831f5cbdf00084671ac

kursawero:
I will take a look.

kursawero:
It looks like using "require" instead of "import ...." solved the issue with the constants, but this could be also just a simple configuration flag of the project that needs to be activated. I will look deeper.

I think the circular dependencies are historic, since we did not really changed the files. For me, it looks like the warning about that will not make the job failing and nobody spotted it before.

kursawero:
It seems that the import statement for the classes Tag.js and DicomMessage.js failed, because the rollup plugin is deprecated ([ see rollupjs troubleshooting](https://rollupjs.org/troubleshooting/#error-name-is-not-exported-by-module).).

There is also a message when you run the build process:

npm WARN deprecated rollup-plugin-babel@4.3.3: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-babel.
npm WARN deprecated rollup-plugin-commonjs@9.3.4: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-commonjs.

It seems that rollup needs to be upgraded too which seems to be not trivial, because you would jump from 1.31.1 to 3.21.0 if you bump to the latest version.

Do you have someone with rollup exprtise in the project? We are using Webpack and usually people with expertise are way faster in adapting the configurations.

IbrahimCSAE:
I believe It's happening because of using module.exports, then trying to import using ES6 syntax in Tag.js. To fix this issue, I believe we can change the export style in dicom.js to ES6.

```JS
export const IMPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2";
export const EXPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2.1";
export const DEFLATED_EXPLICIT_LITTLE_ENDIAN = "1.2.840.10008.1.2.1.99";
export const EXPLICIT_BIG_ENDIAN = "1.2.840.10008.1.2.2";

export const UNDEFINED_LENGTH = 0xffffffff;
export const ITEM_DELIMITATION_LENGTH = 0x00000000;
export const SEQUENCE_DELIMITATION_VALUE = 0x00000000;
```

https://github.com/dcmjs-org/dcmjs/blob/4f1e6754e856d1aca9863def4b82b399bc0c8bb2/src/constants/dicom.js#L1

This will make the error message from rollup go away, I tried building locally, and it worked. You can use import, and you won't need to use require.

You can try this and see if it works now. I also made a pull request if that's easier.

kursawero:
I adapted the pull request.  Thanks for the help.

IbrahimCSAE:
Of course!

sedghi (approved):
Looks good

PR #353, opened 2023-06-19
fix(build): Update build to use newer rollup and fix unit/integration tests

wayfarer3130:
The version of rollup being used was quite old, and there were several failing tests.  Fixed the version of the includes and updated to newer rollup versions.


wayfarer3130:
@pieper  - this is just some output cleanup, plus using newer versions of the libraries.  I did need to switch the unit tests to a different unzip library because there is a fatal bug in the ccurrent version causing the unzip to write corrupted data.  Otherwise there is not actual functional difference.

pieper (approved):

PR #354, opened 2023-06-19
fix: Save units and calibration

wayfarer3130:
This change adds the ability to save the user calibration as well as saving the no-units calibration (there doesn't seem to be a "pixel" units right now)

The proposed format for the user calibration is:
```javascript
{
                RelationshipType: "CONTAINS",
                ValueType: "NUM",
                ConceptNameCodeSequence: {

**This is the code value/meaning for a calibration object, in this case a calibration ruler**
                    CodeValue: "102304005",
                    CodingSchemeDesignator: "SCT",

**Consider the code meaning 'Calibration Ruler'?**
                    CodeMeaning: "Measuring Ruler"
                },
                MeasuredValueSequence: {
                    MeasurementUnitsCodeSequence: - mm measurement,
                    NumericValue: distance entered by the user
                },
                ContentSequence: {
                    RelationshipType: "INFERRED FROM",
                    ValueType: "SCOORD",
                    GraphicType: "POLYLINE",
**Include the image points for the calibration ruler**
                    GraphicData: - points in the image that show the image referenced,
                    ContentSequence: {
                        RelationshipType: "SELECTED FROM",
                        ValueType: "IMAGE",
**The image that is calibrated**
                        ReferencedSOPSequence
                    }
                }
            }
```


wayfarer3130:
@pieper - I added a full set of unit tests to this.  There are a bunch of identical changes to support this, but the actual change is really limited to three changes:
Add Calibration object
Add unit2CodingValue function to determine the correct coding value
Change all the TID300 types to get the right units.

Then, I added unit tests for all the above.

pieper (approved):
Nice - thanks 👍

PR #364, opened 2023-08-24
fix(exports): interpret PN tags as union between object and string

dmlambo:
the JSON standard represents PeopleNames as an object, rather than a delineated string.  this caused SRs derived from metadata to have [object Object] for the patient name, notable in Google's Healthcare dicom store, which seems to pick a random patient name when searching for series. much of the change is attempting to communicate this union somewhat opaquely, and ensuring objects with additional properties aren't destroyed when being naturalized. tested with OHIF master branch.


dmlambo:
A little addendum, you can reproduce the issue in OHIF by creating a single measurement and exporting a structured report. Set a breakpoint [here](https://github.com/OHIF/Viewers/blob/a2ef2b0fcb0ae4a8c02ef07fd97dd6bd6607c509/extensions/default/src/DicomWebDataSource/index.js#L265) (where it does the denaturalizing) and look at the '00100010' (PatientName) tag. It will be "[object Object]".

Finally, this implies a bit of a paradigm shift. In OHIF there are places where it's clear PNs are not handled correcly. [Here, for example.](https://github.com/OHIF/Viewers/blob/a2ef2b0fcb0ae4a8c02ef07fd97dd6bd6607c509/platform/core/src/DICOMWeb/getName.js#L24) With this change, users will have to be careful not to do something like `dataSet.PatientName = "John Doe"` since it will break the accessors. They will have to instead use them, ie, `dataSet.PatientName.Alphabetic = "John Doe"`. Similarly, the accessors should be used for getting the components in a naturalized dataset. I fear this change could affect people in subtle and hard-to-discover ways. 

My change is... kinda sloppy. That's not to say it's sloppily done, rather it reduces cleanliness a fair amount, since it's breaking through some conceptual boundaries. (For instance adding denaturalizing code to ValueRepresentation, sneaking toString and toJSON into native types.) If anyone has a better idea on how to fix it (that doesn't require an epic refactor) I'm all ears.

pieper:
Thanks for working on this 👍 

I'm trying to get my head around the issue - can you elaborate on your comment about derived SR having [object Object] as the Patient Name?

dmlambo:
Sure thing! 

What's happening is when OHIF tries to create a Structured Report, it is pulling instance metadata for the image you are annotating, which is drawn from DICOMWeb as application/dicom+json. The json representation of PN (person name) is not the same as in the binary format (part 10? I'm not super familiar with the naming conventions), as described [here.](https://dicom.nema.org/dicom/2013/output/chtml/part18/sect_F.2.html)

> The Value Field elements of a DICOM attribute with a VR of PN. The non-empty name components of each element are encoded as a JSON strings with the following names:
> - Alphabetic
> - Ideographic
> - Phonetic

Which is to say, when you load a dataset, the Value field of the tag will be different depending on where you load it from. In the case of .dcm, the value will be: 

`data['00100010'].Value = ["Doe^John==doé^yon"]`, 

whereas if loaded from json, it'll be: 

`data['00100010'].Value = [{Alphabetic: "Doe^John", Phonetic: "doé^yon"}]

So, in the latter circumstance, if you were write out the dcm file, eventually PersonName.write() would get called, which is derived from EncodedStringRepresentation, which calls toString() on the Value, and thus it writes out "[object Object]", since it is indeed an object, and not a string. 

This is a problem because OHIF hydrates its structured reports with json data, thus all PN tags become "[object Object]".

A smaller change would be to just detect the type of Value when writing it from the PersonName class, and that would indeed fix binary output, but json output would be broken in the event someone loaded a binary dcm and wrote json. 

I hope it's clearer now!

pieper:
Thanks for the additional context.  This is sounding like more of an OHIF issue than something that should be changed in dcmjs.  The reason I say that is that by design a dataset object in dcmjs should be exactly in DICOM JSON Model, so we should be using the object form with the fields for PersonName and not the ^ separated version used in the Part 10 binary.

Can you point to where in OHIF the SR is being created / used that leads to the reported issue?

In dcmjs, the SR derivation is done by constructing a StructuredReport datasets being referenced as a argument:

https://github.com/dcmjs-org/dcmjs/blob/f844e025f86055a918c13269f6d762a711f3a4bf/src/derivations/StructuredReport.js#L4

This inherits from DerivedDataset, where the PatientName is just copied over:

https://github.com/dcmjs-org/dcmjs/blob/f844e025f86055a918c13269f6d762a711f3a4bf/src/derivations/DerivedDataset.js#L69

dmlambo:
I believe they pull the metadata from [here](https://github.com/cornerstonejs/cornerstone3D/blob/633503fc6db1fac1fcc5992ce5f4d5c1dec42034/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts#L258C42-L258C42) and create the SR just below it, [here](https://github.com/cornerstonejs/cornerstone3D/blob/633503fc6db1fac1fcc5992ce5f4d5c1dec42034/packages/adapters/src/adapters/Cornerstone3D/MeasurementReport.ts#L321).

They also do some [work[1]](https://github.com/OHIF/Viewers/blob/9fbdf37f44a6829a130e9d5bda388f2a4d856556/platform/core/src/utils/formatPN.js#L9) to [detect[2]](https://github.com/OHIF/Viewers/blob/9fbdf37f44a6829a130e9d5bda388f2a4d856556/platform/core/src/DICOMWeb/getName.js#L21) what kind of PN tag they're dealing with.

>  a dataset object in dcmjs should be exactly in DICOM JSON Model

Not according [to the link I posted](https://dicom.nema.org/dicom/2013/output/chtml/part18/sect_F.2.html). I can see, for example, in [the test mock for a minimum viable naturalized dataset](https://github.com/dcmjs-org/dcmjs/blob/f844e025f86055a918c13269f6d762a711f3a4bf/test/mocks/minimal_fields_dataset.json#L6) the name is just a string, which is not what the standard describes. And similarly there is currently no conversion done between the object representation and a plain string of the format "A=B=C" when the dataset is written. Indeed, that's what I'm trying to do with my change.

pieper:
Gotcha, thanks for the links.  It does seem that PN doesn't follow the intended pattern, so that's a bug in the original implementation.

Do your suggested changes alter the API or are both string and object values are now supported?

Also, is it possible to separate the style changes so the functional changes are in a single commit that's easier to review?

dmlambo:
My pleasure! Thanks for the review!

The API hasn't changed, and indeed strings from part10 data and objects from json are supported by the change. 

It should be noted in future documentation that after ingestion (before and after naturalization) one should always use the json accessors to PN tags. After denaturalization, the value in the dict will take on this^string=format. 

I'll separate out the format changes (thanks, lint-on-save 😄 ) and think about adding a couple more tests.

pieper:
thanks again for working on this.  

Does anyone else want to have a look before this gets finalized?  @sedghi maybe?

src/DicomMetaDictionary.js
```diff
@@ -155,11 +155,87 @@ class DicomMetaDictionary {
...
+                    // This following code provides a consistent accessor experience for
+                    // the naturalized dataset. If the value is a string, it mocks a json
+                    // object, and if it's a json object it mocks a string by overriding
+                    // toString. The latter ensures the ValueRepresentation output is
+                    // correct.
+                    if (
+                        data.Value.__pnDcm ||
+                        (data.Value &&
+                            Array.isArray(data.Value) &&
+                            data.Value[0] == "string")
```
wayfarer3130, line 177: I believe you meant typeof data.Value[0] === "string"

src/DicomMetaDictionary.js
```diff
@@ -155,11 +155,87 @@ class DicomMetaDictionary {
...
+                    ) {
+                        data.Value.__objectLike = true;
+                        Object.defineProperty(data.Value, "Alphabetic", {
```
wayfarer3130, line 180: I believe the default encoding is as a List of the typed object, and not as a simple object.  However, making the change on the list itself is probably ok, as it makes it easy to access the data.

src/adapters/Cornerstone/MeasurementReport.js
```diff
@@ -319,9 +317,8 @@ export default class MeasurementReport {
                   );
 
             if (toolClass) {
-                const measurement = toolClass.getMeasurementData(
-                    measurementGroup
-                );
+                const measurement =
```
wayfarer3130, line 320: Seems like a lot of formatting changes - if you want to run the formatter, do it as a separate PR that is just that change so that the review only has the substantive changes.

test/DicomMetaDictionary.test.js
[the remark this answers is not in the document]
pieper: @dmlambo did already split the PR into 3 commits so you should be able to review those independently.  Maybe you are right though @wayfarer3130 and we should think whether it should be a different PR.  Is there a practical benefit down the line to having two PRs?

src/DicomMetaDictionary.js
dmlambo: Indeed!

src/DicomMetaDictionary.js
dmlambo: I thought given it was a special case we may as well enforce this "objectlike" behaviour.

dmlambo:
Any more comments? 🙂

src/utilities/dicomJson.js
```diff
@@ -0,0 +1,127 @@
+/**
+ * Converts a naturalized PersonName (json) to a part10 PersonName (string)
+ * @param {object} value json-style person name
+ * @returns {string} part10 dicom representation of PersonName
+ */
+function pnDenaturalize(value) {
+    if (value && typeof value === "object") {
+        return `${value.Alphabetic ?? ""}=${value.Ideographic ?? ""}=${
```
wayfarer3130, line 8: It should only get the ideographic/phonetic separators if one of them exists.  That is:

if( value.Ideographic || value.Phonetic ) return your full string
return value.Alphabetic

That avoids you creating a string with the = in it when it isn't necessary (which changes the appearance in some systems).

src/utilities/dicomJson.js
```diff
@@ -0,0 +1,127 @@
...
+
+    // This following code provides a consistent accessor experience for
+    // the naturalized dataset. If the value is a string, it mocks a json
+    // object, and if it's a json object it mocks a string by overriding
+    // toString. The latter ensures the ValueRepresentation output is
+    // correct.
+    if (
+        data.Value.__pnDcm ||
+        (data.Value &&
+            Array.isArray(data.Value) &&
```
wayfarer3130, line 64: This doesn't deal with tags such as ConsultingPhysician's Name which can contain 1...n instances.  I know that is really a pain to have to deal with, as it will have to be decided how that works, but my suggestion is to:
1. If the VM !== 1 or 0 or 1, then leave the object as an array of objects all the time.
2. For other VM sizes, apply the change to the individual strings in each item, as a new String(originalValue)

src/anonymizer.js
```diff
@@ -259,13 +259,15 @@ export function cleanTags(
             var tagNumber = tagInfo.tag,
                 tagString = Tag.fromPString(tagNumber).toCleanString();
             if (dict[tagString]) {
-                var newValue;
-                if (tagString in tagNamesToReplace) {
-                    newValue = [tagNamesToReplace[tagString]];
+                if (Array.isArray(dict[tagString].Value)) {
```
wayfarer3130, line 262: I don't think this is the right behaviour because it modifies the original string, which is going to cause problems.  You do NOT want to keep the references here because of that.  Also, you don't want to clear every single array element there is, only the ones which have tag names to replace.  The other ones should be left alone.

src/utilities/dicomJson.js
```diff
@@ -0,0 +1,127 @@
...
+    // To ensure proper json output, this value will be treated as the json dicom
+    // equivalent when serializing.
+    tag.values.toJSON = function () {
+        const components = this[0]?.split("=");
+        if (components && components.length) {
+            return [
+                {
+                    Alphabetic: components[0],
+                    Ideographic: components[1],
+                    Phonetic: components[2]
```
wayfarer3130, line 32: This needs to be a map for all the elements of the array in case the vm>1.   What I would do is check if the base value is a string, and if it is, then convert it into an array and perform the mapping function on the result.  Probably having a function like toArray = (value) => Array.isArray(value) ? value : [value]
You might need to undefined  check as well.

src/utilities/dicomJson.js
```diff
@@ -0,0 +1,127 @@
...
+    // object, and if it's a json object it mocks a string by overriding
+    // toString. The latter ensures the ValueRepresentation output is
+    // correct.
+    if (
+        data.Value.__pnDcm ||
+        (data.Value &&
+            Array.isArray(data.Value) &&
+            typeof data.Value[0] == "string")
+    ) {
+        data.Value.__objectLike = true;
```
wayfarer3130, line 67: Can you make this a hidden property so it doesn't show up on iteration?

test/anonymizer.test.js
wayfarer3130, line 1: Given the changes to the anonyization above, please add a test with a sequence (array) of values that is not anonymized, and ensure that sequence is unchanged in the original object after you anonymize it.

test/sample-sr.json
wayfarer3130, line 1: Could you add a multiple name example here please?  That allows you to add the appropriate tests to handle multiple names.

src/utilities/dicomJson.js
dmlambo: Notice the .replace just below which removes all end-of-string adjacent =. Thus if the full string is "ABC==" it becomes "ABC", "ABC=DEF=" becomes "ABC=DEF", and "ABC==GHI" stays the same.

src/utilities/dicomJson.js
dmlambo: Oh my. I indeed didn't know PN could have VM > 1. I'll do another pass with that in mind, thanks for the heads up.

dmlambo:
Incoming. Heh, this one is rather large, though some changes were reverted from before.

Note that as it stands one test will fail, because multiplicity.dcm does not have the correct tags, and I don't have access to the test data repo. Let me know where to send it to get it released!

This change may take a little while to explain and to understand, so take your time looking through it, and I'll answer any questions. 

Basically whenever a tag root gets created (in the form of `{ vr: vr, values: values }` or `naturalDataset`) or when a tag itself gets assigned (the `values` or `Values` properties, as well as any natural name property) some accessors will get placed on the object to allow one value to be represented in another way, by way of toJSON and toString. It's a similar concept to the previous changes, but wider scoped. 

For instance: 
```
    static naturalizeDataset(dataset) {
        const naturalDataset = ValueRepresentation.addTagAccessors({
            _vrMap: {}
        });
```
which replaces the dataset with a Proxy which intercepts any property assignment, looks up the propert's ValueRepresentation, and sets additional accessors on the value:
``` 
                set(target, prop, value) {
                    var vrType;
                    if (
                        ["values", "Value"].includes(prop) &&
                        target.vr &&
                        ValueRepresentation.hasValueAccessors(target.vr)
                    ) {
                        vrType = ValueRepresentation.createByTypeString(
                            target.vr
                        );
                    } else if (prop in DicomMetaDictionary.nameMap) {
                        vrType = ValueRepresentation.createByTypeString(
                            DicomMetaDictionary.nameMap[prop].vr
                        );
                    } else {
                        target[prop] = value;
                        return true;
                    }

                    target[prop] = vrType.addValueAccessors(value);

                    return true;
                }
```
This is to ensure if someone assigns a PN incorrectly (IE, without using the json model), we can still output correct JSON, as well as allowing us to convert to part10 given the same scenario. 

I'll likely have a few smaller changes coming through, but let me know what you think in the meantime.

test/sample-sr.json
dmlambo: added it to multiplicity.dcm, but it requires an update of the file in the test artifacts repo.

dmlambo:
In the end other changes worked out in the wash, it seems pretty stable to me. Does anyone have any more comments on this change? Anything I can explain better? Thanks!

pieper:
Thanks for your work on this @dmlambo 👍 

From a read-through I believe I'm fine with this but would appreciate if @wayfarer3130 can look at the unresolved conversations.

Also I see a test failure now: https://github.com/dcmjs-org/dcmjs/actions/runs/6178725114/job/17111489516?pr=364

dmlambo:
> Also I see a test failure now: https://github.com/dcmjs-org/dcmjs/actions/runs/6178725114/job/17111489516?pr=364

This is because the multiplicity dcm in the test data repo doesn't have my changes. I've added a PN tag with multiplicity. I don't have permissions to update that file, so someone else would have to do it for me. 🙂 

[multiplicity.zip](https://github.com/dcmjs-org/dcmjs/files/12737838/multiplicity.zip)

Or if someone wants to modify the existing file, I used [DICOM Editor Tool](https://www.dvtk.org/dicom/editor/) to add this tag:

<HTML>
<BODY>
<TABLE><TR><TD>(0008,1070)</TD><TD>Operators' Name</TD><TD>PN</TD><TD>2</TD><TD>Doe^John\Doe^Jane</TD></TR></TABLE>

</BODY>
</HTML>

Thanks!

src/ValueRepresentation.js
```diff
@@ -41,6 +49,68 @@ class ValueRepresentation {
...
+            //   ...
+            //   jsonOutput = JSON.serialize(dicomDict);
+            // or:
+            //   naturalizedDataset.OperatorsName = 'Doe^John';
+            //   jsonOutput = JSON.serialize(naturalizedDataset);
+            // Whereas the correct usage of the dicom+json model would be:
+            //   dicomDict.dict.upsertTag('00101001', 'PN', [{Alphabetic:'Doe^John'}]);
+            //   naturalizedDataset.OperatorsName = [{Alphabetic:'Doe^John'},{Alphabetic:'Doe^Jane'}];
+            // TODO: refactor with addAccessors.js in mind
+            const handler = {
```
wayfarer3130, line 83: Could you define the const handler statically?  Creating a new instance of it every time isn't needed, as the values inside it are static.

src/ValueRepresentation.js
```diff
@@ -670,41 +764,85 @@ class PersonName extends EncodedStringRepresentation {
     constructor() {
         super("PN");
         this.maxLength = null;
-        this.padByte = 0x20;
+        this.padByte = PADDING_SPACE;
+    }
+
+    static checkComponentLengths(components) {
```
wayfarer3130, line 770: I see value in not writing non-conformant JSON, but reading it should be fully supported - can you test reading non-conformant data and then indicate here that this function only prevents one from creating it?

test/anonymizer.test.js
```diff
@@ -24,13 +24,65 @@ it("test_anonymization", () => {
...
+        JSON.stringify([
+            {
+                Alphabetic: "Doe^John",
+                Ideographic: "Johnny",
+                Phonetic: "Jonny"
+            },
+            { Alphabetic: "Doe^Jane", Ideographic: "Janie", Phonetic: "Jayne" }
+        ])
+    );
+    expect(otherPatientNamesIDValue.toString()).toEqual(
```
wayfarer3130, line 68: Thank you for adding tests here for the patient name handling - that is definitely worthwhile having this tested.

wayfarer3130:
This is looking good now - much happier with it as updated.  Could you move the handler out to a single const rather than redeclaring it in the creator?  Then I'm happy with it.

pieper:
> > Also I see a test failure now: https://github.com/dcmjs-org/dcmjs/actions/runs/6178725114/job/17111489516?pr=364
> 
> This is because the multiplicity dcm in the test data repo doesn't have my changes. I've added a PN tag with multiplicity. I don't have permissions to update that file, so someone else would have to do it for me. 🙂
> 
> [multiplicity.zip](https://github.com/dcmjs-org/dcmjs/files/12737838/multiplicity.zip)
> 
> Or if someone wants to modify the existing file, I used [DICOM Editor Tool](https://www.dvtk.org/dicom/editor/) to add this tag:
> 
> (0008,1070)	Operators' Name	PN	2	Doe^John\Doe^Jane
> Thanks!

I'm happy to put new testing data in the repository, but do you really need to change and existing file?  Wouldn't it be better to add a new file and leave existing tests/data in place?

pieper:
Okay, data published as an asset here: https://github.com/dcmjs-org/data/releases/tag/multiplicity2

src/ValueRepresentation.js
dmlambo: As it stands it's only checked on write. I'll add a comment about it.

dmlambo:
Rebased on top of master. Should be good to go unless there are any more notes I missed! Thanks for all the feedback everyone!

test/data.test.js
```diff
@@ -532,18 +542,31 @@ it("test_custom_dictionary", () => {
...
-    const dcmPath = await getTestDataset(url, "multiplicity.dcm")
+    const dcmPath = await getTestDataset(url, "multiplicity.dcm");
     const file = await promisify(fs.readFile)(dcmPath);
     const dicomDict = DicomMessage.readFile(file.buffer);
 
     expect(dicomDict.dict["00101020"].Value).toEqual([1, 2]);
     expect(dicomDict.dict["0018100B"].Value).toEqual(["1.2", "3.4"]);
 });
 
+it("Reads DICOM with PersonName multiplicity", async () => {
```
wayfarer3130, line 553: Thanks for testing this.

wayfarer3130 (approved):

pieper:
@wayfarer3130 is this ready to merge?  Sounds like it's needed for https://github.com/OHIF/Viewers/pull/3739

@sedghi @jbocce we can merge this as far as I'm concerned.

sedghi:
Looks like good to go

wayfarer3130:
I had approved the PR, yes, I think this is good to go.

Issue #366, opened 2023-10-02
Length of Decimal String larger than 16 characters


rvantklooster:
Given a Decimal String with the numerical value of 0.99990081787109 (Number), and writing it to a DICOM file, then the resulting string is 17 characters wide.

In DecimalString::formatValue (ValueRepresentation.js), the function toExponential() guarantee that the resulting string is maximum 16 characters wide. See https://github.com/dcmjs-org/dcmjs/blob/master/src/ValueRepresentation.js#L535C1-L545C6

We copied the behavior from pydicom to properly truncate the string and made a PR. See https://github.com/pydicom/pydicom/blob/main/src/pydicom/valuerep.py#L912

PR #367, opened 2023-10-02
Fixed issue with Decimal String conversion resulting in more than 16 …

rvantklooster:
Fixed issue with Decimal String conversion resulting in more than 16 characters and added unit test.
See https://github.com/dcmjs-org/dcmjs/issues/366

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -533,23 +533,35 @@ class DecimalString extends AsciiStringRepresentation {
     }
 
     formatValue(value) {
-        if (value === null) {
-            return "";
-        }
+        if (value === null) return "";
+
         let str = String(value);
         if (str.length > this.maxLength) {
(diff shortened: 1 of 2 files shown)


pieper (approved):
Makes sense to me.

@swederik you thumbed up the issue, can you take a look at this before we merge just for another set of eyes?  Do you think there should be a comment explaining what the code does?

swederik:
Yes completely agree, the code is good but a short comment about why it exists will be helpful for the next person who reads it. Especially happy that @rvantklooster added a test! Thanks!

swederik (approved):
Looks good but please add a short comment above the logic explaining why it is there, thanks!

rvantklooster:
I updated the documentation (in line with the documentation of its pydicom counterpart). Nice to see this quick turnaround!

PR #352, opened 2023-06-05
Process UN VR with DicomDictionary

jimOnAir:
Hello. I have an issue with receiving images from [KPacs](https://image-systems.biz/products/free-dicom-pacs-tools/k-pacs/) with dcmjs-dimse library:

```
value.toExponential is not a function : TypeError: value.toExponential is not a function >>     at DecimalString.formatValue
```

After some investigation I have found that KPacs replaces VR of some tags with UN.
Original file:

```
(0002,0000) UL FileMetaInformationGroupLength = 176
(0002,0001) OB FileMetaInformationVersion = <bin: 0x0001>
(0002,0002) UI MediaStorageSOPClassUID = 1.2.840.10008.5.1.4.1.1.1.1
(0002,0003) UI MediaStorageSOPInstanceUID = 1.3.6.1.4.1.18047.1.6.1005961685460676437
(0002,0010) UI TransferSyntaxUID = 1.2.840.10008.1.2.4.70
(0002,0012) UI ImplementationClassUID = 1.2.40.0.13.1.3
(0002,0013) SH ImplementationVersionName = dcm4che-5.17.0
(0008,0005) CS SpecificCharacterSet = ISO_IR 100
(0008,0008) CS ImageType = ORIGINAL\PRIMARY
(0008,0016) UI SOPClassUID = 1.2.840.10008.5.1.4.1.1.1.1
(0008,0018) UI SOPInstanceUID = 1.3.6.1.4.1.18047.1.6.1005961685460676437
(0008,0020) DA StudyDate = 19000101
(0008,0021) DA SeriesDate = 19000101
(0008,0022) DA AcquisitionDate = 19000101
(0008,0023) DA ContentDate = 19000101
(0008,0030) TM StudyTime = 111111
(0008,0031) TM SeriesTime = 111111
(0008,0032) TM AcquisitionTime = 111111
(0008,0033) TM ContentTime = 111111
(0008,0050) SH AccessionNumber = AccNum_1
(0008,0060) CS Modality = DX
(0008,0068) CS PresentationIntentType = FOR PRESENTATION
(0008,0070) LO Manufacturer =
(0008,0080) LO InstitutionName =
(0008,0090) PN ReferringPhysicianName =
(0008,1010) SH StationName =
(0008,1030) LO StudyDescription = ELBOW LT
(0008,103E) LO SeriesDescription = ELBOW LT AP
(0008,1040) LO InstitutionalDepartmentName =
(0008,1050) PN PerformingPhysicianName =
[...]

PR #379, opened 2024-02-13
Fix UN VR setting during naturalization/denaturalization

andreidubov:
Hello, I encountered several issues with saving the dataset to a file. This is also related to this PR #352 . 
Some of them: 
```
Value exceeds max length, vr: CS, value: [object ArrayBuffer], length: undefined
value.toExponential is not a function
```
After investigation I found that the source file has UN VR for some tags. And the main problem is that after the `naturalizeDataset` process we lose UN VR for these tags and in the `denaturalizeDataset` process we get not UN VR, but VR according to the dictionary. New VR does not match the value and therefore we get the following errors when trying to write.

Tag with UN VR in file 
image

This tag un dict after `DicomMessage.readFile`
image

This tag in dataset after `DicomMetaDictionary.naturalizeDataset`
image

This tag after `DicomMetaDictionary.denaturalizeDataset` (DS VR instaed of UN)
image

Solution: 
~~Saving the UN VR to the `_vrMap` in dataset and use it in the `denaturalizeDataset` to restore the original UN VR.~~
Based on #352 PR. If the VR is UN and does not match the VR in dictionary for this tag, then parse the tag's value according to the VR from the dictionary. This logic is controlled by the `parseUnknownVr` flag, which is disabled by default.


jimOnAir:
@pieper, @PantelisGeorgiadis could you please have a look on this?

jimOnAir:
Thanks for review. 
It is the same case I had with KPacs. It changes VR of all unknown tags to UN.
At least this change allows us to save images to file as it was received. Previously there were errors while trying to process ArrayBuffer as number or other type.

pieper:
Why doesn't kpacs know about ExposureIndex if it's in the standard?  Maybe it needs an updated data dictionary?

jimOnAir:
It doesn't know. And all unknown tags are coded with UN VR. KPacs is not supported for at least decade. But some organisations are still using it.

pieper:
I'm not saying this is not important, but workarounds for out of date and unsupported external software should be isolated from the main logic of the package.  The reason for this is that many dicom toolkits have historically become cluttered with code that was only needed for data of a particular nonstandard type.  Sometimes these workaround are not well commented or rely on data that can't be shared for privacy reasons or can't be reproduced because they require some commercial installation.  For these reasons I'd like the library itself to focus mainly on very standard dicom and move special case handling somewhere else.

In this case I'd suggest adding a very specific filter for code that is known to require interoperability with kpacs where it applies a specific standardization step converting UN to DS for this tag (and maybe makes similar fixes for other tags with known VRs).

src/DicomMetaDictionary.js
```diff
@@ -126,6 +126,10 @@ class DicomMetaDictionary {
                     // when the vr is data-dependent, keep track of the original type
                     naturalDataset._vrMap[naturalName] = data.vr;
                 }
+
+                if (data.vr == "UN") {
```
wayfarer3130, line 130: I agree with @pieper that this should be a separate step that should be configurable - maybe as an option going into the parse step, but also as a stand alone filter that works on any JSON DICOMweb metadata instance.  That filter should then have some way to register named fixes and a way to disable fixes individually.

andreidubov: Could you please clarify this point, I want to make sure that I understood you correctly. Do you mean that  I should add an option(flag) to `naturalizeDataset/denaturalizeDataset` and depending on this option(flag) apply or skip for this logic?

pieper: I was thinking you could have a special-purpose piece of code at the application level that is coded based on the unique information that it will be accepting data from a source (kpacs) that has known limitations in terms of generating standard dicom objects.  I.e. in this case you know that certain tag values will have incorrect values.  This special-purpose code should read the instance and apply whatever fixes are needed to make the instance standards-compliant and then send it to the dcmjs code.

We may want to provide a standard place for this kind of special-purpose code to happen.  Perhaps in dcmjs-dimse there could be a hook to apply a standardizing transform to the incoming instance data.  If there end up being very common ones they could be bundled as add-on packages (e.g. a set of transforms for kpacs, ones for non-standard microCT scanners, etc).

jimOnAir: We don't have access to this low-level data on the application level as the application provides C-Store service only thing we have from dcmjs-dimse is Dataset without any VR information. dcmjs-dimse relies on dcmjs while reading the Dataset. In this case, we don't know if something is wrong until we have read the data.
What you propose is rewriting parts of dcmjs and dcmjs-dimse that are receiving binary data over the network and naturalizing the dataset. 
The other thing is that we don't have access to the KPacs dictionary and can't compare it to dcmjs Dictionary to find out which tags can have the wrong VRs.

pieper: From what I understand if dcmjs-dimse is processing a c-store you should be able to access the instance as it is received.  From there you can read it as pure dicom json model and you can look at each of the tags in the instance and see if they have the non-standard form that leads to the issue based on the dicom dictionary and your knowledge of the deficiencies with kpacs.  (That is, if you see that a tag with a know standard VR has been set to UN by kpacs).  If you see an issue, you can modify the instances and save them back before further processing.

andreidubov: At the application level in CStore we do not have DICOM json, we only have naturalized Dataset. At this level, there is no way how we could get the original VR for tag.
image

DICOM json object is present inside dsmjs-dmse in the buffer/file reading functions.
image
image

If we talk about saving the original UN VR and saving it to a file without getting an error (as suggested in PR). I think it’s impossible to do without changes in dcmjs. The denaturalizeDataset takes the VR based only on the value in the dictionary, which leads to a mismatch between the VR and the value. If we talk about converting UN VR value to VR from the dictionary, it looks like a more difficult task to handle all possible VRs. @PantelisGeorgiadis do you have any ideas how this issue could be solved at the dcmjs-dimse level?

PantelisGeorgiadis:
Hello all!
Sorry for being silent all that time! Well, dcmjs-dimse was designed having simplicity in mind while trying to abstract the calling code from “advanced DICOM internals”. However, as it is getting more adopted, it is obvious that users need more flexibility so they can handle more complex and demanding cases. 

In that direction, we have created this PR [https://github.com/PantelisGeorgiadis/dcmjs-dimse/pull/41](https://github.com/PantelisGeorgiadis/dcmjs-dimse/pull/41m), a long time ago (!!!), that it allows the reception of DICOM content to Writable streams. It actually allows the receiving code to create an object inherited from `Writable`, that will receive the incoming bytes from the network (in chunks), giving the freedom to perform any operation on them before the `cStoreRequest` event is called. A major use case is to perform streaming to disk while receiving big instances.

There are still a few technical issues to be solved (e.g. handling of drain stream event) but I believe that it fits your needs, since it will allow you to access the incoming bytes and apply a “custom-parsing” logic. Thoughts?

andreidubov:
Hello everyone. I changed the logic based on #352  PR. Now If the VR is UN and does not match the VR in dictionary for this tag, then parse the tag's value according to the VR from the dictionary (we will have the correct value corresponding to VR from the dictionary and there will be no errors when writing). Also I added `parseUnknownVr` option which is disabled by default. Maybe this solution would be better. What do you think?

andreidubov:
Hi @pieper, @wayfarer3130  could you please have a look on this?

pieper:
Can you clarify if you think that KPACS behavior is actually correct according to the dicom standard?  If it doesn't know about the tag because it's dictionary is out of date is it correct for it to set the VR to UN?  And if that's the case, then do we know for sure that since we have a more recent dictionary then parsing the value with the correct VR is the supposed to work?  If that's the case then I don't think we need to have a flag for this as it should always be done (with error checks).

If on the other hand this is invalid DICOM then I think something like what @PantelisGeorgiadis suggested is more appropriate.

andreidubov:
As @jimOnAir wrote [here](https://github.com/dcmjs-org/dcmjs/pull/379#issuecomment-1948287493), it is correct to set the VR to UN if this tag is not in application dictionary. 

- _An existing application that does not support that new Attribute (and encounters it) could convert the VR to UN._ [https://dicom.nema.org/](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.2.html)
image

- It also says that we can parse an element with UN VR if we know its VR from our dictionary: _If at some point an application knows the actual VR for a Data Element of VR UN (e.g., has its own applicable data dictionary), it can assume that the Value Field of the Data Element is encoded in Little Endian byte ordering with Implicit VR encoding, irrespective of the current Transfer Syntax._
[https://dicom.nema.org/](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.2.html)
image

pieper:
@wayfarer3130 @PantelisGeorgiadis do you agree this PR now represent standard behavior?  Do you see any cases where it will won't be backwards compatible?

@andreidubov @jimOnAir thanks for bearing with the process.

One more request: we try to keep binary files out of the repo.  Can you redo the test to download the file from here: https://github.com/dcmjs-org/data/releases/tag/unknown-VR

You can copy how it's done in other tests.

wayfarer3130:
> @wayfarer3130 @PantelisGeorgiadis do you agree this PR now represent standard behavior? Do you see any cases where it will won't be backwards compatible?
> 
> @andreidubov @jimOnAir thanks for bearing with the process.
> 
> One more request: we try to keep binary files out of the repo. Can you redo the test to download the file from here: https://github.com/dcmjs-org/data/releases/tag/unknown-VR
> 
> You can copy how it's done in other tests.

Yes, I agree that it now looks like it is standards compliant

pieper:
Thanks @wayfarer3130 

@andreidubov if you update the test I'll be fine with merging this.

PantelisGeorgiadis:
Thank you @andreidubov and @jimOnAir for working on this! I agree that it is fine merging this, after updating the test.

andreidubov:
Thanks guys, sorry for the wait, test updated.

pieper:
Thanks for updating the test 👍 

I see there's an unrelated test failing, so I created an issue for it https://github.com/dcmjs-org/dcmjs/issues/380

pieper:
@PantelisGeorgiadis do you want to create a new issue related to the PN changes you mentioned?

PR #384, opened 2024-03-14
feat/nm normalizer

ahlaughland:
This PR adds a normalizer for Nuclear medicine (NM) images. This allows saving a SEG file via the normalizer (from OHIF in our case).  

This code was tested locally using our current version of OHIF (3.8.0-beta.52). The SEG file seen below was created in OHIF, saved via the NM normalizer code and loaded back into OHIF:
image


pieper:
Looks reasonable to me - thanks for working on this 👍 

I see you're still adding commits, so let me know when this is ready to go.

PR #385, opened 2024-03-18
Update package version to 0.32.0

ahlaughland:
I missed this small detail when submitting my PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dcmjs",
-  "version": "0.31.0",
+  "version": "0.32.0",
   "description": "Javascript implementation of DICOM manipulation",
   "main": "build/dcmjs.js",
   "module": "build/dcmjs.es.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dcmjs",
-  "version": "0.30.0",
+  "version": "0.31.0",
   "description": "Javascript implementation of DICOM manipulation",
   "main": "build/dcmjs.js",
   "module": "build/dcmjs.es.js",


ahlaughland:
Maybe this should be 0.32.0 since this PR will update to v0.32.0

pieper:
We haven't been consistent on this version number, rather relying on the semantic commit approach that updates the npm package versions.  Seems like people just update now-and-then.

@ahlaughland If you know or can find out the correct policy for how this number should be managed it would be great to add a note [to the readme](https://github.com/dcmjs-org/dcmjs/blob/master/README.md?plain=1#L75-L87) so we can be consistent going forward.  Thanks 👍

ahlaughland:
> We haven't been consistent on this version number, rather relying on the semantic commit approach that updates the npm package versions. Seems like people just update now-and-then.
> 
> @ahlaughland If you know or can find out the correct policy for how this number should be managed it would be great to add a note [to the readme](https://github.com/dcmjs-org/dcmjs/blob/master/README.md?plain=1#L75-L87) so we can be consistent going forward. Thanks 👍

I will take a look and see what I can do. Thanks again!

ahlaughland:
@pieper I looked at a few repos and the `version` in `package.json` is typically updated as part of the PR. This is how we do it but mostly not. 😄 

OHIF for example, removed `version` from their `package.json` and moved it to a `version.json` file https://github.com/rapidllc/3drd-ohif-v3/blob/main/version.json.

The only recommendation I could make is to manually update `version` in the DCMJS `package.json` as part of PR approvals. Other than that the process looks good.

PR #391, opened 2024-05-06
fix: Publish and build package on newer node

wayfarer3130:
Updated a number of versions, and started fixing the circular dependency issues to try to get the overall build/release working.
This does change the node version used to build, as well as the rollup version being used, but is intended to be entirely non-functional, just changing things trying to get the main build working.
@pieper  - this PR is just more than a basic test to try to release things again, so I'd like you to take a look before I push it.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -15,9 +15,9 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Node.js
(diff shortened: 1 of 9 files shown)


pieper (approved):
Thanks for working on this 👍 

I think those changes all look reasonable.

pieper:
Now that the deploys are working it's time to actually fix the examples.  There are CORS issues and some other things.

PR #392, opened 2024-05-06
feat: Netlify app master-dcmjs2 deploy build

wayfarer3130:
The change uses the URL that built correctly on the last deploy of dcmjs.  Looks like it should work this time.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Setup Node.js
         uses: actions/setup-node@v1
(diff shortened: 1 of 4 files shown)


pieper:
Thanks @wayfarer3130 - I'm not sure what 'lury' refers to, but it seems to be something @swederik and @JamesAPetts added so maybe they know?

https://github.com/dcmjs-org/dcmjs/commit/0e1e2921e953bde8f4707b025f584a6c558ffedd

The vtk.js / dicom seg example uses some data of me from IsomicsPublic bucket and the CORS is set correctly there, but there's been some change in dicomweb-client that breaks the ArrayBuffer parsing (it's now an Array with one ArrayBuffer element, where the code is expecting an ArrayBuffer directly - I'm not sure when that behavior came in but it should be easy to fix).  I guess now I can just test a fix with the web editor and test it using the netlfy preview - cool!

pieper (approved):

wayfarer3130:
@pieper - had to fix a race condition on the download as well so that the failing test now passes consistently.

PR #393, opened 2024-05-06
fix(vtkseg): fix part10 parsing

pieper:


pieper:
Shoot, it looks like there's some other problem at the vtk.js level with this example.  It's pretty old and now it's getting these errors:

`WebGL: INVALID_OPERATION: texImage3D: ArrayBufferView not big enough for request`

I don't have time to look into this now which is too bad because this is a good demo.

wayfarer3130 (approved):

PR #394, opened 2024-05-06
test: Race condition test

wayfarer3130:
Here is a proper fix for the race condition in the test data download


pieper (approved):
Sounds good 👍

PR #395, opened 2024-06-07
fix(Segmentation): Fix only one frame

jlopes90:


simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

pieper:
Thanks for investigating this and providing the PR with a link to the cornerstone issue.  While that link is helpful, please make your contribution here self-contained and describe in your commit message, and in code comments as needed, specifically, please clarify:
* what the issue is you are trying to fix
* what the behavior is without this change
* what the behavior is after this change
* why this change fixes the issue
* why you don't think this will have any side-effects for other code in the future

It would also be great to have a unit test that illustrates this.  That is, a test that demonstrates a valid use-case that fails without this change but works with the change.

Also, see the guidelines for commit messages and correct your commit as indicated:
https://github.com/dcmjs-org/dcmjs?tab=readme-ov-file#for-maintainers-and-contributors

I don't mean to be super-picky here, but this repo is used by a lot of people and subtle bugs can occur if we aren't careful.  I want to be sure someone who sees your contribution in the future fully understands it.

jlopes90:
`generateSegmentation` supports two or more frames and does not support just one frame and gives an error.

When testing just one frame for example:
```typescript
Cornerstone3D.Segmentation.generateSegmentation([cacheImage], labelmapData, metaData);
```
and shows the results of errors.


## First Error

```typescript
let distance1 = distanceDatasetPairs[1][0];
```

image
image

So I created a new variable, `distance1` is zero, for just one frame. But if `NumberOfFrames` was two or more get this `distanceDatasetPairs[1][0]`.

```typescript
// If the frame is one, it is zero
let distance1 = 0;

if (ds.NumberOfFrames > 1) {
    distance1 = distanceDatasetPairs[1][0];
}
```

## Second Error

```typescript
if (ds.NumberOfFrames === 1)
    ds.PerFrameFunctionalGroupsSequence = [
        ds.PerFrameFunctionalGroupsSequence
    ];
```

image
image
image

I removed this code and it works similar to two or more frames.

jlopes90:
There is another way without using this pull request or change.


```javascript
How does just one frame work:
Cornerstone3D.Segmentation.generateSegmentation(
    [cacheImage, cacheImage],
    labelmapData,
    metaData
);
```

jlopes90:
There is another way without using this pull request or change.

How does just one frame work:
```javascript
Cornerstone3D.Segmentation.generateSegmentation(
    [cacheImage, cacheImage],
    labelmapData,
    metaData
);
```

`[cacheImage, cacheImage]` both "cacheImages" are the same, just ignore it and it works only one frame.

PR #399, opened 2024-08-12
Other double as explicit value representation

craigberry1:
Add reserved 2 byte padding to `OtherDouble` read logic and read value length as 32 bit integer [per spec](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/chapter_7.html). 

Specifically section 7.1.2 that states:
`
for VRs of AE, AS, AT, CS, DA, DS, DT, FL, FD, IS, LO, LT, PN, SH, SL, SS, ST, TM, UI, UL and US the Value Length Field is the 16-bit unsigned integer following the two byte VR Field. The value of the Value Length Field shall equal the length of the Value Field.`

Meaning `OD` should be treated as:
`for all other VRs the 16 bits following the two byte VR Field are reserved for use by later versions of the DICOM Standard. These reserved bytes shall be set to 0000H and shall not be used or decoded. The Value Length Field is a 32-bit unsigned integer.
`


craigberry1:
Not doing this causes data elements after the first `OD` tag to be corrupted. This means reading and writing a file with `OD` will not produce a valid DICOM file:

image

test/OD-single-word-example.dcm
craigberry1, line 1: This is a synthetic and simple DICOM image that has SOP Instance UID , SOP Class UID, and the two tags in my unit test.

Putting this PR out to get initial feedback and if things look good I will transition this asset to https://github.com/dcmjs-org/data/ and refactor the test to pull from there.

@pieper might ask for some guidance on exactly how to create the release with my test file.

pieper:
Thanks for working on this 👍 

Even though it's a small file, let's put it in as a data release asset for consistency.  Can you update the test, remove the dcm file and update the PR?

https://github.com/dcmjs-org/data/releases/tag/od-encoding-data

craigberry1:
> Thanks for working on this 👍
> 
> Even though it's a small file, let's put it in as a data release asset for consistency. Can you update the test, remove the dcm file and update the PR?
> 
> https://github.com/dcmjs-org/data/releases/tag/od-encoding-data

Done 👍

Issue #398, opened 2024-08-05
Loss of precision when serializing Decimal String (DS) and Integer String (IS) 


craigberry1:
When reading a DICOM file, data elements with VR `IS` or `DS` are currently serialized by reading as ASCII Strings, applying custom formatting, and then converting to `Number`.

Examples:
- `1.0000` drops padded zeroes and yields `1`
- `+1.234` drops the `+` sign
- `9007199254740993` is read as `9007199254740992` due to floating point precision
- `       456    ` drops whitespace to `456`

My [read from the spec](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html#:~:text=2%20bytes%20fixed.-,DS,-Decimal%20String) is that these are all considered valid values:
```
A string of characters representing either a fixed point number or a floating point number. A fixed point number shall contain only the characters 0-9 with an optional leading "+" or "-" and an optional "." to mark the decimal point. A floating point number shall be conveyed as defined in ANSI X3.9, with an "E" or "e" to indicate the start of the exponent. Decimal Strings may be padded with leading or trailing spaces. Embedded spaces are not allowed.
```

This leads to fundamental changes in the data element value by simply reading and then writing a DICOM file.

Example

When reading in a file containing the data elements:
```
(0x0018,0x1041) DS Contrast/Bolus Volume 	 VR=<DS>   VL=<0x0010>  <9007199254740993>
(0x0018,0x1190) DS Focal Spot(s) 	 VR=<DS>   VL=<0x0006>  <00.000>
```

A read & write of it using dcm-js with no other modifications yields:
```
(0x0018,0x1041) DS Contrast/Bolus Volume 	 VR=<DS>   VL=<0x0010>  <9007199254740992> // modified value
(0x0018,0x1190) DS Focal Spot(s) 	 VR=<DS>   VL=<0x0002>  <0 > // loss of significant digits
``` 

Assuming that these are in fact valid values according to the spec, why does the library try to apply custom formatting to the value like dropping whitespace and the `+` sign, or bother converting to a `Number` which is subject to Floating Point rounding issues? Why not leave as a String and let clients decide when/how to parse as a number?

pieper:
There's been discussion of this in the past.  See: https://github.com/dcmjs-org/dcmjs/issues/53, https://github.com/dcmjs-org/dcmjs/pull/97, https://github.com/dcmjs-org/dcmjs/pull/99

[...]

PR #400, opened 2024-08-16
feat(lossless-round-trip): Implement lossless read writes

craigberry1:
Solves https://github.com/dcmjs-org/dcmjs/issues/398 by implementing lossless round trip read/writes to preserve original file data when tags are unmodified

- Extracts any formatting or manipulation done to data element values into `ValueRepresentation.applyFormatting`
- Maintain reference to original value via new `_rawValue` property on a data element
- Detect changes to `Value` during writes to determine whether to maintain original unformatted value or update
- Filter `BinaryValueRepresentation` from lossless processing by default to avoid duplicating large arrayBuffers like pixel data, but provide override mechanism via `readOptions`

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -1334,30 +1334,21 @@ class ParsedUnknownValue extends BinaryRepresentation {
         const streamFromBuffer = new ReadBufferStream(arrayBuffer, true);
         const vr = ValueRepresentation.createByTypeString(this.type);
 
-        var values = [];
         if (vr.isBinary() && length > vr.maxLength && !vr.noMultiple) {
+            var values = [];
+            var rawValues = [];
             var times = length / vr.maxLength,
                 i = 0;
-            while (i++ < times) {
-                values.push(vr.read(streamFromBuffer, vr.maxLength, syntax));
-            }
-        } else {
-            var val = vr.read(streamFromBuffer, length, syntax);
-            if (!vr.isBinary() && singleVRs.indexOf(vr.type) == -1) {
-                values = val;
-                if (typeof val === "string") {
-                    values = val.split(String.fromCharCode(VM_DELIMITER));
-                }
-            } else if (vr.type == "SQ") {
-                values = val;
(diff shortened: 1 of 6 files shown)


test/data.test.js
```diff
@@ -357,7 +357,9 @@ it("test_exponential_notation", () => {
     dataset.ImagePositionPatient[2] = 7.1945578383e-5;
     const buffer = data.write();
     const copy = dcmjs.data.DicomMessage.readFile(buffer);
-    expect(JSON.stringify(data)).toEqual(JSON.stringify(copy));
```
pieper, line 360: Did we lose this test somehow?

src/ValueRepresentation.js
```diff
@@ -79,6 +79,7 @@ class ValueRepresentation {
         this._allowMultiple =
             !this._isBinary && singleVRs.indexOf(this.type) == -1;
         this._isExplicit = explicitVRs.indexOf(this.type) != -1;
+        this._storeRaw = true;
```
pieper, line 82: I was wondering if this should be false in the general case to save space.  It seems to me that a very common use case is to read and operate on the data without planning to store it out again.  But unless I'm missing something this overhead is going to be very small compared to the PixelData it's really in the noise level so I'm okay with this being true by default.

pieper (approved):
Thanks for working on this 👍 

Other than the one possible missing test it looks good to me and we can merge.

test/data.test.js
craigberry1: This did fail after my changes but it's behaving as expected since the test modifies the dataset and then re-writes it (`_rawValue` on `ImagePositionPatient` no longer matches since `Value` was modified prior to write).

So instead of expecting exact equality between original/new I just verify `ImagePositionPatient` tag. Based on the name `"test_exponential_notation"` I think that was the main spirit of this test.

src/ValueRepresentation.js
craigberry1: I'd be fine making it a optional `readOption` if you prefer. But yes space impact should be minimal as it's disabled for binary VRs and sequences.

PR #401, opened 2024-08-26
fix(lossles-round-trip): Account for padding byte in numeric strings with multiplicity

craigberry1:
Fixes issue during lossless read/write with numeric string value representations (`IS`/`DS`). This was caused by a string like:
`0.99924236548978\-0.0322633220972\-0.0217663285287\0.02949870928067\0.99267261121054\-0.1171789789306 ` where the space following the final number is correctly added for evenness, but produced an error during re-write as the unformatted value exceeded the max length of the VR.

Solution was to remove this padding byte from `_rawValue` property during read if it exceeds the max length of the vr. This padding will be re-added by the writer if needed for evenness.

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -65,6 +65,19 @@ function toWindows(inputArray, size) {
     );
 }
 
+// TODO: Move into value representation?
+export function dropPadByte(values, vr) {
+    if (values.length > 1) {
+        const last = values[values.length - 1];
+
+        // trim padding byte if last element exceeds length by exactly 1
+        if (last.length === vr.maxLength + 1 && last.charAt(vr.maxLength) === String.fromCharCode(vr.padByte)) {
+            values[values.length - 1] = last.substring(0, vr.maxLength);
+        }
+    }
+    return values;
+}
+
 let DicomMessage, Tag;
 
 var binaryVRs = ["FL", "FD", "SL", "SS", "UL", "US", "AT"],
@@ -318,6 +331,10 @@ class AsciiStringRepresentation extends ValueRepresentation {
 
     readBytes(stream, length) {
         return stream.readAsciiString(length);
+    }
+
+    applyFormatting(value) {
+        return value.trim();
     }
 
     writeBytes(stream, value, writeOptions) {
@@ -604,7 +621,7 @@ class CodeString extends AsciiStringRepresentation {
 
     readBytes(stream, length) {
         const BACKSLASH = String.fromCharCode(VM_DELIMITER);
-        return stream.readAsciiString(length).split(BACKSLASH);
+        return dropPadByte(stream.readAsciiString(length).split(BACKSLASH), this);
     }
 
     applyFormatting(value) {
@@ -654,7 +671,7 @@ class AttributeTag extends ValueRepresentation {
 class DateValue extends AsciiStringRepresentation {
     constructor(value) {
         super("DA", value);
-        this.maxLength = 18;
+        this.maxLength = 8;
         this.padByte = PADDING_SPACE;
         //this.fixed = true;
         this.defaultValue = "";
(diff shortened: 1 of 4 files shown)


craigberry1:
This can be an issue for other VRs as well. Converting to draft while coming up with a more generalized solution.

craigberry1:
Fixed generically by add `dropPadByte` check when strings are split for multiplicity. Aims to track specific case when the last element of an multiple values data element has is exactly 1 byte over max length and that byte is the VR's pad byte. In this case we drop that character as it is not considered part of the original value but was just added for evenness. The writer will add the padding back if needed.

pieper (commented):
This looks good to me, thanks for the work and the PR 👍 

I'd like if we could get a second set of eyes on it, just to be sure Iso 'll request another reviewer to have a look.

src/ValueRepresentation.js
```diff
@@ -136,6 +136,42 @@ class ValueRepresentation {
         return tag;
     }
 
+    /**
+     * Removes padding byte, if it exists, from the last value in a multiple-value data element.
+     *
+     * This function ensures that data elements with multiple values maintain their integrity for lossless
```
wayfarer3130, line 142: I think the pad byte is an artifact of the binary encoding, and should always be stripped - that makes it consistent as to what value one gets.  Depending on the maximum character length means one ends up getting different results depending on the exact boundary/parse conditions which is a recipe for errors and bugs.

wayfarer3130:
> This looks good to me, thanks for the work and the PR 👍
> 
> I'd like if we could get a second set of eyes on it, just to be sure Iso 'll request another reviewer to have a look.

I think always removing the pad byte from numeric strings is a more consistent solution than trying to get the exact correct length.  It just seems like that leaves the possibility of errors when the length is that exact length as compared to shorter lengths or vice-versa.

craigberry1:
> > This looks good to me, thanks for the work and the PR 👍
> > I'd like if we could get a second set of eyes on it, just to be sure Iso 'll request another reviewer to have a look.
> 
> I think always removing the pad byte from numeric strings is a more consistent solution than trying to get the exact correct length. It just seems like that leaves the possibility of errors when the length is that exact length as compared to shorter lengths or vice-versa.

The issue I had was trying to determine whether the pad byte/space was significant or added for evenness. Many VR's (like the numeric strings) allow for whitespace so unless the parsed string exceeds the max length there is nothing saying the whitespace shouldn't be there.

This originally come out of work for [lossless round trip processing](https://github.com/dcmjs-org/dcmjs/issues/398#issuecomment-2269286677) where the goal has been to preserve exactly what is in the source DICOM file. 

With that approach in mind my thinking was to only manipulate the original string in a very specific circumstance, in this case being the final element in the string containing a single byte longer than allowed and it being equal to the padding byte. In any other case I leave as-is and let the writer determine if the value is valid during write.

wayfarer3130:
@craigberry1 - as long as you only remove the pad byte on the last item when it is at an odd position, you should be fine as it will then get re-added.  I agree the standard is ambiguous there, but I don't quite see how to fix it at this point.

craigberry1:
> @craigberry1 - as long as you only remove the pad byte on the last item when it is at an odd position, you should be fine as it will then get re-added. I agree the standard is ambiguous there, but I don't quite see how to fix it at this point.

Hey @wayfarer3130 thanks for the response, I was away for a few days. 

What do you think of the current implementation? I go one step further to determine when the padding byte was added by looking specifically for whitespace at "maxLength + 1". 

```javascript
// If the last element exceeds the max length by exactly one character and ends with the padding byte,
// trim the padding byte to avoid a max length violation during write.
if (lastValue.length === maxLength + 1 && lastValue.endsWith(padChar)) {
    values[lastIdx] = lastValue.substring(0, maxLength); // Trim the padding byte
}
```

With a tag like `DS` `0.123\ +0.00 `, if I strip the pad byte just due to odd length it will convert ` +0.00 ` -> ` +0.00`. The writer will re-add the space to achieve evenness but we have stripped the padding where it may have been intended (preserving this was [the goal](https://github.com/dcmjs-org/dcmjs/issues/398#issuecomment-2269286677) with the `_rawValue` property).

I can be more aggressive in stripping the padding byte if you think its the right call though. The writer will add it back for even length so the lossless round trip feature still works.

craigberry1:
> @craigberry1 - as long as you only remove the pad byte on the last item when it is at an odd position, you should be fine as it will then get re-added. I agree the standard is ambiguous there, but I don't quite see how to fix it at this point.

Thanks for the response @wayfarer3130, I was away for a few days. I wrote up another comment but deleted it after some more thought because I think the odd length check you suggested is sufficient in this case. I just want to be careful not to regress on the lossless round trip work.

I'll update this PR shortly. Thanks for the help.

craigberry1:
@pieper @wayfarer3130 Ready for another look when you get a chance. Thanks!

pieper:
I think that only stripping a space at the end of the string is the right thing.  I don't think any reasonable implementation would add a byte anywhere else to make the buffer and even length.  Sounds like we are all agreed then.

PR #402, opened 2024-09-18
Add OCT normalizer

igorsimko:
This PR adds support for Ophthalmic Tomography Image modality. 

Fixes also https://github.com/OHIF/Viewers/issues/4293#issuecomment-2340923886

PR #403, opened 2024-09-18
feat: Add OP normalizer

igorsimko:
This PR adds support for Ophthalmic Photography 8 Bit Image modality. Often time referenced by OCT (https://github.com/dcmjs-org/dcmjs/pull/402)


igorsimko:
@pieper I went over OP normalizer again and I don't fully understand whether it's needed or not. Problem in this bug report refers to only OCT scan when trying to export segmentation for it. I haven't encountered segmentations for OP modality yet. Segmentations are usually only visualized on OP from OCT. 

See following example where A is OCT and B is OP
image


OP file I attached (from realt device) doesn't have bunch of dicom tags which are used in `normalizer.js` so I would have to just use simple

```
class OPImageNormalizer extends Normalizer {
    normalize() {
        this.dataset = this.datasets[0];
    }
}
```

Do you think we should still add OP to normalizers in that case?

pieper:
I'm not familiar enough with OCT and OP to know for sure, but as a general design pattern for dcmjs it's good to have normalizers for any sop class so that code can rely on them providing a fixed up version of any non-standard dicom (e.g. fix up any vendor screwups).  Many sop class normalizers can be basically no-ops, which is good.

igorsimko:
Thank you for you feedback. In that case I updated PR to also include test and run simple normalization as written in comment above.

sedghi (approved):
Thanks, since test is passing i'm gonna merge it, great work

igorsimko:
@sedghi that's great to hear, thanks a lot! Shall we do same with https://github.com/dcmjs-org/dcmjs/pull/402?

PR #405, opened 2024-10-15
fix(package.json): Address MPL by replacing logging lib to a MIT licensed one

igoroctaviano:
Use MIT logging lib instead of MPL

- Old library: https://github.com/shellscape/loglevelnext/blob/master/src/LogLevel.ts
- Updated readme 
- Updated tests


pieper (approved):
Thanks for the fix 👍

PR #406, opened 2024-10-15
fix(src/test): Address lint issues + setup lint/format CI action

igoroctaviano:
- Address lint issues
- Add format and ci lint action to avoid merging broken/unformatted code

### DIFF
--- a/src/sr/templates.js
+++ b/src/sr/templates.js
@@ -1,32 +1,22 @@
 import { Code, CodedConcept } from "./coding.js";
 import {
     CodeContentItem,
-    CompositeContentItem,
     ContainerContentItem,
     ContentSequence,
-    GraphicTypes,
-    GraphicTypes3D,
-    ImageContentItem,
     NumContentItem,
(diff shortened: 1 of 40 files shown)


src/sr/templates.js
```diff
@@ -302,7 +292,7 @@ class MeasurementStatisticalProperties extends Template {
                 value: options.authority,
                 relationshipType: RelationshipTypes.HAS_PROPERTIES
             });
-            this.push(authorityItem);
+            this.push(descriptionItem);
```
pieper, line 295: Looks like this was a bug before and isn't part of the code formatting?

pieper (commented):
Looks great overall.  Just not sure about that one line.

src/sr/templates.js
igoroctaviano: True but I catched it through lint

pieper: Cool!  Good catch 👍

PR #414, opened 2024-12-18
feat(vr): Save original non-standard VR

alekschernof:
Hello, I encountered several issues with saving the dataset to a file. 
After investigation I found that the source file has different vr that  in dictionary And the main problem is that after the `naturalizeDataset` process we lose origin vr for these tags and in the `denaturalizeDataset `process we try to process tag with incorrect vr 
and therefore we get the following errors when trying to write.

image

In dicom image (LUTData tag has incorrect vr. In standart it US)

Screenshot from 2024-12-18 15-09-51

We received that image and we get the following errors when trying to write

Screenshot from 2024-12-18 12-23-12

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -126,6 +126,10 @@ class DicomMetaDictionary {
                     // when the vr is data-dependent, keep track of the original type
                     naturalDataset._vrMap[naturalName] = data.vr;
                 }
+                if (data.vr !== entry.vr) {
+                    // save origin vr if it different that in dictionary
+                    naturalDataset._vrMap[naturalName] = data.vr;
+                }
             }
 
             if (data.Value === undefined) {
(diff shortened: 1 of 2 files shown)


pieper:
This sounds reasonable, but a couple things.

First, we need to have commits follow the [commit message format](https://github.com/dcmjs-org/dcmjs?tab=readme-ov-file#for-maintainers-and-contributors).  Here it would be something like `feat(vr): Save original non-standard VR` and also including a summary of the change in the rest of the commit message so that it's explained in  `git log` without needing to refer back to the PR.

Also, please add a test for this. It can be simple, just make a dataset from scratch that has the incorrect VR and then run it through the round trip (go from namified javascript object to part 10, then back from part10 to namified object and back to part 10 and then back to namified and confirm that desired VR has been preserved).

alekschernof:
> This sounds reasonable, but a couple things.
> 
> First, we need to have commits follow the [commit message format](https://github.com/dcmjs-org/dcmjs?tab=readme-ov-file#for-maintainers-and-contributors). Here it would be something like `feat(vr): Save original non-standard VR` and also including a summary of the change in the rest of the commit message so that it's explained in `git log` without needing to refer back to the PR.
> 
> Also, please add a test for this. It can be simple, just make a dataset from scratch that has the incorrect VR and then run it through the round trip (go from namified javascript object to part 10, then back from part10 to namified object and back to part 10 and then back to namified and confirm that desired VR has been preserved).

Hello, I renamed pr and commits by requirements and added simple test for `naturalizeDataset` and `denaturalizeDataset` with checks

pieper (approved):
Thanks LGTM 👍

PR #419, opened 2025-01-15
Preserve optional DICOM SegmentDescription during renumbering

rogerbramon:
This pull request adds support for handling the optional DICOM SegmentDescription field during segment renumbering. If the source segment includes a SegmentDescription, it will be copied to the renumbered segment.


rogerbramon:
@pieper, could you please review this PR and let me know your thoughts?

pieper (approved):
LGTM 👍

PR #420, opened 2025-01-17
fix(dicomMessage): to handle undefined return value

sedghi:
I have some data that fails to load in OHIF when draged and dropped ([try](https://viewer-dev.ohif.org/localbasic) ), and i have narrowed down the issue to an undefined value which this PR fixes

Here is the issue which has the data attached

https://github.com/OHIF/Viewers/issues/4588#issuecomment-2598032763

--

At OHIF, we use namify, and apparently, this data contains some undefined values that need to be addressed during the process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "build": "rollup -c",
     "build:examples": "npm run build && npx cpx 'build/**/*.{js,map}' examples/js",
     "start": "rollup -c -w",
+    "dev": "rollup -c -w",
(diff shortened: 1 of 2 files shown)


wayfarer3130 (approved):

pieper:
Can you include a brief description of why this change was causing the OHIF issue you cited?

sedghi:
@pieper i updated the description, it is during the namify process

Issue #418, opened 2025-01-02
Tag consistency issue during roundtrip file loading/saving/loading


PantelisGeorgiadis:
Trying to update the `dcmjs` version used in `dcmjs-dimse` beyond v0.34.0, I noticed that two of the `dcmjs-dimse`'s unit tests are failing. These tests (actually they test around the same area but in a sync/async way) are performing a roundtrip loading/saving/loading of an encapsulated PDF fixture file and check whether all elements are in place. 

The tests are passing for `dcmjs` versions lower than v0.34.0 and, after some debugging, I narrowed down the behavior-changing pull request to be the "Implement lossless read writes #400".

The DICOM PDF file contains an empty Referring Physician Name (0008,0090) tag (zero-lenght value) which when loaded and naturalized, it ends up as `ReferringPhysicianName: undefined`. When this object is getting denaturalized, due to the fact that Referring Physician Name's in undefined, the whole element is omitted from the dataset.

The different behavior is that versions prior v0.34.0 are loading this tag as an empty array and, therefore, when the elements are written back to a file, the tag isn't omitted.

I used the attached DICOM PDF file and the following code to verify the above.
[pdf.dcm.zip](https://github.com/user-attachments/files/18290439/pdf.dcm.zip)
```js
const { readFileSync, writeFileSync } = require('fs');
const path = require('path');

const dcmjs = require('dcmjs');
const { DicomDict, DicomMessage, DicomMetaDictionary } = dcmjs.data;

function loadP10Buffer(path, readOptions) {
  readOptions = readOptions || {};
  const baseOptions = { ignoreErrors: true };
  readOptions = { ...baseOptions, ...readOptions };

  const buffer = readFileSync(path);
  const dicomDict = DicomMessage.readFile(
    buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength),
    readOptions
  );
  const meta = DicomMetaDictionary.naturalizeDataset(dicomDict.meta);
  const elements = DicomMetaDictionary.naturalizeDataset(dicomDict.dict);

  return { meta, elements };
}

function saveP10Buffer(path, meta, elements, writeOptions) {
  const denaturalizedMetaHeader = DicomMetaDictionary.denaturalizeDataset(meta);
  const dicomDict = new DicomDict(denaturalizedMetaHeader);
  dicomDict.dict = DicomMetaDictionary.denaturalizeDataset(elements);

[...]

Issue #417, opened 2024-12-30
Tag value [undefined] causing unable to write a dicomDict


salimkanoun:
Hi there,

I have some DICOMs causing some errors during write with DcmJs,

My code is pretty simple, it read a dicom using 
` const originalDicomDict = dcmjs.data.DicomMessage.readFile(arrayBuffer);`

Modify some tags according to a custom logic and then rewrite it using

`originalDicomDict.write({ allowInvalidVRLength: true });`

However I found some DICOM in which the write() function throws an error : 
`dcmjs.js?v=61259bcc:4670 Uncaught (in promise) Error: Not a number: undefined
    at toInt (dcmjs.js?v=61259bcc:4670:15)
    at WriteBufferStream2.writeUint16 (dcmjs.js?v=61259bcc:4728:46)
    at dcmjs.js?v=61259bcc:61805:50
    at Array.forEach (<anonymous>)
    at UnsignedShort2.write (dcmjs.js?v=61259bcc:61800:38)
    at UnsignedShort2.writeBytes (dcmjs.js?v=61259bcc:62884:177)
    at Tag2.write (dcmjs.js?v=61259bcc:63225:34)
    at dcmjs.js?v=61259bcc:63494:32
    at Array.forEach (<anonymous>)
    at Function.write (dcmjs.js?v=61259bcc:63487:24)`

After investigation, these error is thrown by tags having `Value : [undefined]` (array length of 1 with an single undefined value), I found this error into the "001021C0" and "00189337" but may also affect other tags eventually.

I'm not sure if it is a read error or a write error, the Value : [undefined] seems a bit curious, maybe a read error generating a Value : [undefined] instead of Value : [] ?

You can find here two instances generating the bug  : https://drive.google.com/drive/folders/1rM5YwuEXFSe52YL_mcqmpP_9wQRmuMtL?usp=sharing

in the dicom_error_dcmjs.DCM => see the tag 00180012 => 00189337
in the dicom_error_dcmjs_2.dcm => see the tag 001021C0

pieper:
Thanks for reporting this, we want dcmjs to be as robust as possible. That said, I suspect this issue is the result of a combination of some issues with the source data and possibly with the custom logic you are using.  For example, the "_2" can't be read with dcmdump from dcmtk.

What would be best if you could contribute a test that demonstrates the issue so it's easy to reproduce.

[...]

PR #421, opened 2025-02-02
fix(lossless-round-trip): Support naturalized PN with empty tag value 

craigberry1:
Issue was found where a 0 length `PN` tag would be removed from a DICOM file following naturalization and re-encoding.

This was due to lossless round trip implementation (https://github.com/dcmjs-org/dcmjs/pull/400) which was accounting for tag multiplicity in `PN` during `ValueRepresentation.read`, however the `dicomJson` serializer/deserializer was also [accounting for multiplicity](https://github.com/dcmjs-org/dcmjs/blob/master/src/utilities/dicomJson.js#L18) during its conversion.

This led to a subtle problem where the `tag.Value` property was `[undefined]`, which naturalized to `undefined`, which was then removed from the dataset altogether by denaturalize. 

Fixes this issue by treating PersonName `_rawValue` property as a String with multiplicity still retained and allows `dicomJson` to do the splitting. 

Fixes https://github.com/dcmjs-org/dcmjs/issues/418


craigberry1:
Added fix for https://github.com/dcmjs-org/dcmjs/issues/417 where `ValueRepresentation.read` did not return expected object structure `{ _rawValue, Value }` for empty + fixed VRs.

@pieper I added raw DCM files from the two respective issues mentioned here but I know we are moving towards the dcmjs-data repo. So just let me know how you want me to proceed here with those assets.

Thanks

pieper:
Thanks for the fix 👍 (sorry I was out last week and just catching up now)

I do want to avoid having dcm files in the repo so I added them here: https://github.com/dcmjs-org/data/releases/tag/empty-tag-round-trip. It would be great if you could tweak the PR to download them as part of the test.

craigberry1:
Hey @pieper welcome back. I will get to this later today.

craigberry1:
Should be all set now @pieper

pieper:
Thanks @craigberry1 but I think you will need to force-push to remove the dcm files from the PR?

craigberry1:
Ah, yes you are correct. Removed and pushed.

pieper:
Thanks 👍

PR #426, opened 2025-03-12
fix(value-representation): adjust DA, DT, and TM maxLength for query ranges

mishijima:
Hello.

Thank you for having this package available for the public to use.

I'd like to report an issue where the `StudyDate` value is being truncated when it contains query ranges because it exceeds the maximum length.

For example, the `StudyDate` value is truncated when it exceeds eight characters. As a result, when specifying a date range (e.g., `StudyDate: '20010101-20250301'`), only the start date (`20010101`) is retained.

Prior to [this commit](https://github.com/dcmjs-org/dcmjs/commit/7b44953a5e6d1b4fb4377a427b77435dc443252e#diff-dab0f010af348fdebdba9b603d33a2e371af5662ae4f11612f3e57e3e7c6d572R692), the maximum length was set to 18, which aligns with the [DICOM standard - 6.2 Value Representation (VR)](https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html)

Thank you to @PantelisGeorgiadis for suggesting that I create a pull request in this repository. Please refer to the [original issue](https://github.com/PantelisGeorgiadis/dcmjs-dimse/issues/70#issue-2913711665) that I created in the `dcmjs-dimse` repository.

This is my first time creating a pull request in this repository, so I apologise in advance if I have missed anything due to my limited knowledge.

The changes apply for Date (DA), Date Time (DT), and Time (TM).

References:
- https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html
- https://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_c.2.2.2.5.html


pieper (approved):
LGTM 👍

pieper:
Looks good to me - thanks for the contribution 👍

mishijima:
Much appreciated for the review @pieper 👍

mishijima:
Hello @pieper,

Apologies for the oversight in my recent contribution. Upon further review, I realise that adding the `writeBytes` function for `DateValue`, `DateTime`, and `TimeValue` was incorrect and unnecesary.

Do you think it's okay to set the maxLength of `DateValue`, `DateTime`, and `TimeValue` to the maximum length of the matching range query, so when it doesn't get truncated in https://github.com/dcmjs-org/dcmjs/blob/37c41c67b5244e9f5354d00bfecd1ae70a90cd29/src/DicomMetaDictionary.js#L273-L277

Thanks for your understanding, and I will work on submitting a fix for this.

pieper:
From what I read I think it would make sense to add the check for the `'-'` character here, so that you set the longer max length when it's a range.  Do you agree?

mishijima:
> From what I read I think it would make sense to add the check for the `'-'` character here, so that you set the longer max length when it's a range. Do you agree?

Hello @pieper,

Thank you for your reply and understanding.

I think it makes sense to handle it here since this is where the value gets truncated. For attributes that support Range Matching, we can use a different maximum length.

I will look into it.

PR #429, opened 2025-04-17
feat: add text box group to TID300 measurements

pedrokohler:
This PR adds the ability to store the position of a textbox or label within an SR.

With this enhancement, libraries such as OHIF and Cornerstone can save and reload the textbox positions that users manually set.

Currently, this isn't possible, as textbox positions are lost when saving the SR.

### DIFF
--- a/src/utilities/TID300/TID300Measurement.js
+++ b/src/utilities/TID300/TID300Measurement.js
@@ -12,8 +12,40 @@ export default class TID300Measurement {
             ...this.getTrackingGroups(),
             ...this.getFindingGroup(),
             ...this.getFindingSiteGroups(),
-            ...contentSequenceEntries
+            ...contentSequenceEntries,
+            ...this.getTextBoxGroups()
         ];
+    }
+
+    getTextBoxGroups() {
+        const { textBoxPoint } = this.props;
+
+        if (
+            !textBoxPoint ||
+            typeof textBoxPoint.x !== "number" ||
+            typeof textBoxPoint.y !== "number"
+        ) {
+            return [];
+        }
+
+        const textBoxSequence = {
+            RelationshipType: "CONTAINS",
+            ValueType: "SCOORD",
+            ConceptNameCodeSequence: {
+                CodeValue: "111030",
+                CodingSchemeDesignator: "DCM",
+                CodeMeaning: "Annotation Position"
+            },
+            GraphicType: "POINT",
+            GraphicData: [textBoxPoint.x, textBoxPoint.y],
+            ContentSequence: {
+                RelationshipType: "SELECTED FROM",
+                ValueType: "IMAGE",
+                ReferencedSOPSequence: this.ReferencedSOPSequence
+            }
+        };
+
+        return [textBoxSequence];
     }
 
     getTrackingGroups() {


pieper:
Hi @pedrokohler - I missed the last coordination meeting, but did you run this by David and Andrey?

pedrokohler:
> Hi @pedrokohler - I missed the last coordination meeting, but did you run this by David and Andrey?

Hi @pieper, this isn't related to IDC. A client wants to add functionality to save the textbox positions of annotations in their OHIF fork, and they'd like to contribute this back to the open-source project.

However, to achieve this, we also need to make a few changes in both Cornerstone and dcmjs.

pieper:
Thanks Pedro.  I ask because we want to keep the dcmjs implementation as standards-compliant as possible and it seems that [code 111030 is "Image Region"](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_d.html#:~:text=111030,Image%20Region) and not "Annotation Position".  I believe the code meaning text is meant to follow the standard exactly and not to encode different information.

I understand the need to serialize the text position though.  It may be that there is no standard way to do that.  If that's the case, I suggest that Cornerstone / OHIF will probably need to just add private tags to the top level of the SR describing the annotation positions and maybe other settings that don't have a standardized DICOM representation.

pedrokohler:
@pieper I see. Sorry I didn't double check the code meaning.

I understand your concern and I agree. But maybe we can find a valid alternative within the standard.

Checking the link you provided I was able to find this:

image

Do you think we can use it? Maybe @wayfarer3130 can also come up with suggestions.

pieper:
I guess using an annotation node would be sort-of okay, but also maybe not really keeping with the standard since those are probably meant to be human readable.  Plus I'm sure once people get started they will want to save things like the font size or color in addition to the position.  In the extreme people could put long xml strings or something.

I think a toolkit-specific set of private tags are unfortunately the only think I know can be used here in spirit of the standard.  Yes, if @wayfarer3130 knows better, that would be great.  We can also raise this to David C or Chis B since maintaining dcmjs/cs3d/ohif dicom correctness is within the IDC mission.

wayfarer3130:
Actually I think if we included hte annotation note text itself, as well as the point data it would probably be closer to the intention of the standard than the hack that is currently used to store custom annotation notes in OHIF and Microscopy viewer.  However, I'd like to pass this by David C. and see what he suggests.

pedrokohler:
Let me try tagging him here @dclunie

pieper:
I don't think `@` tagging David C will trigger an email since he hasn't been involved in this thread.  

I put this topic on the agenda to discuss with him at next Tuesday's IDC meeting.   If you need something sooner we can try emailing him directly.

@wayfarer3130 what microscopy hack are you referring to?

pedrokohler:
@wayfarer3130 

image

I think you mistakenly updated @pieper's comment.

wayfarer3130:
> I think you mistakenly updated @pieper's comment.
Yes, it does look like it got edited rather than replied to.  I'm not sure how to restore the rest of the original comment.

wayfarer3130:
> Presentation states are intended for this - see https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.10.5.html

That presupposes that one is willing to have multiple representations of annotations, and to split the data across multiple files.  The SR format is a single contained format for creating structured annotations for sharing as annotations/measurements.   Presentation states are really about the display of the annotation data.

The existing SR measurements have a way to store the position of the annotation, eg by storing the start and end points of a caliper length measurement.   They also have a good way to store finding codes and the finding site codes.  I'm not sure how they should store freeform text that is associated with the finding/measurement.  Currently OHIF uses a custom finding code whose value is the freeform text - that seems like a bad idea to me.   Also, that text can be moved around, and we'd like to have a way to store the position it gets moved to within the single SR file so that all the data for the measurements is stored in one file rather than trying to split and correlate them across multiple files.

wayfarer3130:
@pieper - What about adding $QualType and $ImagePurpose arrays from https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1501
to the return type for the measurement report/tid 1500 generation code.  That is according to the DICOM standard.  Then, the specified purpose of $ImagePurpose is to specify what the measurements apply to. The purpose of the QualType is:
"Allows encoding a flat list of name-value pairs that are coded questions with coded or text answers, for example, to record categorical observations related to the subject of the measurement group. A single level of coded modifiers may be present."

So, I believe that second example is exactly what the cornerstone free text is intended to apply to, while the first one can be used for other purposes, but one of the possible imagePurpose is to specify a possible location on an image for text.  It is only possible, nad can be used in other ways.

Examples:
 How to specify the volume
This specifies an image contained in the volume of interest, with $ImagePurpose
```
            {
              "RelationshipType": "CONTAINS",
              "ConceptNameCodeSequence": {
                "CodeValue": "ImageVolumeSelector",
                "CodingSchemeDesignator": "99CORNERSTONE"
              },
              "ValueType": "IMAGE",
              "ReferencedSOPSequence": {
                "ReferencedSOPClassUID": "1.2.840.10008.5.1.4.1.1.7",
                "ReferencedSOPInstanceUID": "2.25.146740561248393523358491776463474639750"
              }
            },
```

The right way to specify freeform text location associated with this measurement group  $ImagePurpose entry
```
            {
              "RelationshipType": "CONTAINS",
              "ConceptNameCodeSequence": {
                "CodeValue": "FreeformText",
                "CodingSchemeDesignator": "99CORNERSTONE",
                "CodeMeaning": "Freeform text"
              },
              "ValueType": "TEXT",
              "TextValue": "This is where the standard text qualitative evaluations should go"
            },
```

One way to specify freeform text location associated with this measurement group $ImagePurpose
```
            {
              "RelationshipType": "CONTAINS",
              "ConceptNameCodeSequence": {
                "CodeValue": "FreeformText",
                "CodingSchemeDesignator": "99CORNERSTONE",
                "CodeMeaning": "Freeform text"
              },
              "ValueType": "SCOORD",
              "GraphicType": "POLYLINE",
              "GraphicData": [
                770.6092529296875, 422.8263854980469, 942.9541625976562, 280.6865539550781
              ],
              "ContentSequence": {
                "RelationshipType": "SELECTED FROM",
                "ValueType": "IMAGE",
                "ReferencedSOPSequence": {
                  "ReferencedSOPClassUID": "1.2.840.10008.5.1.4.1.1.7",
                  "ReferencedSOPInstanceUID": "2.25.146740561248393523358491776463474639750"
                }
              }
            }
```

fedorov:
Did I understand it correctly that you are suggesting to use the combination of TID1501 rows 10c (to encode location of the text box) and 11 (to assign text to be displayed)?

I don't think this could ever be interoperable, but I agree this may be a way to (ab)use the template to achieve your goal, without creating an invalid object. I guess this would be equivalent of using private tags but at the level of SR content tree.

pieper:
I don't have a strong preference of methods.  As long as you are using a 99 coding scheme and publish the meaning somewhere (perhaps in github and include the URL in the SR somewhere)  there's a good chance people (or AIs) will be able to sort out what was intended.

wayfarer3130:
> Did I understand it correctly that you are suggesting to use the combination of TID1501 rows 10c (to encode location of the text box) and 11 (to assign text to be displayed)?

Yes, that is what I'm proposing.
 
> I don't think this could ever be interoperable, but I agree this may be a way to (ab)use the template to achieve your goal, without creating an invalid object. I guess this would be equivalent of using private tags but at the level of SR content tree.

It might be interoperable if we get agreement from IHE, perhaps as part of the AIRAI standard - I have a contact working on that right now, so I'm going to talk with her about adding a couple of things to the proposal.

wayfarer3130:
> I agree it is better to do this than using private tags (at least with private coding schemes, the meaning is available within the instance, one one won't need to chase down the non-existent conformance statement to figure it out!
> 
> But you should make sure in all cases you have the complete `CodeMeaning, CodeValue, CodingSchemeDesignator` tuple, and also I think it is a pretty bad practice to use the same content for `CodeMeaning` and `CodeValue` - instead it is better to assign some alphanumeric string and have a system of how to add new values to the coding scheme. It would be best if you maintained the list of `99CORNERSTONE` private codes prominently, ideally as part of the repo, which I would think should be this repo, which in turn raises the question why private coding scheme is `99CORNERSTONE` and not `99DCMJS`?

It was mostly a quick example of roughly what I was thinking about rather than the exact final version.
For sure, full complete tag values, and an additional way of defining the code values as an exported file within DCMJS so that values can be shared across different implementations.  Ideally if we can actually get a DICOM CP to add the annotations as standard values with examples in the IHE AIRAI upcoming standard to ensure interoperability.

wayfarer3130:
@pedrokohler - lets close this PR for now as it can be done entirely on the generating side, and then think about how to generalize the support for annotations in dcmjs once we have some examples in Cornerstone on how to handle it.  THe existing support is sufficient to be able to generate these examples, and it can be discussed from there what part should be created upstream.

PR #427, opened 2025-03-21
fix(value-representation): prevent truncation of DA, DT, and TM in denormalization

mishijima:
### Follow-up to [#426](https://github.com/dcmjs-org/dcmjs/pull/426)  

This PR addresses the following:  

- Reverts changes related to `writeBytes` for `DA`, `DT`, and `TM`.  
- Adds `allowRangeMatching` to the `ValueRepresentation` class to indicate whether a VR supports range matching.  
- Updates the truncation logic in `denaturalizeDataset`:  
  - If `vr.allowRangeMatching()` and `vr.rangeMatchingMaxLength` return `true`, `maxLength` is set to `vr.rangeMatchingMaxLength`.

PR #433, opened 2025-05-03
Using `checkLength` to determine max length for DA, DT, TM (Query with range matching)

PantelisGeorgiadis:
This PR patches the work of @mishijima on #427.

More specifically,
- Removes the allowRangeMatching because the existence of rangeMatchingMaxLength, for a VR, gives the same information.
- Uses the existing checkLength function for DA, TM and DT to determine a range query and return whether the length is valid or not.
- Updates the unit tests to perform a full roundtrip between the naturalized and denaturalized worlds.
- The build output was used with dcmjs-dimse to make sure that range queries are working as expected.

@pieper please review and if you approve and merge, please close #427 since it is redundant.

### DIFF
--- a/src/DicomMetaDictionary.js
+++ b/src/DicomMetaDictionary.js
@@ -270,11 +270,16 @@ class DicomMetaDictionary {
 
                     if (!vr.isBinary() && vr.maxLength) {
                         dataItem.Value = dataItem.Value.map(value => {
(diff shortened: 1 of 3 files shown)


PantelisGeorgiadis:
Thank you for merging this PR @pieper! Any plans for releasing this?

PR #435, opened 2025-05-06
Update templates.js

### LINKED ISSUE #434: TrackingIdentifier constructor calls super(options) and injects raw options into the ContentSequence
### Reproduction Steps

1. Draw one or more ROIs in the [Slim viewer](https://github.com/ImagingDataCommons/slim) (each gets a generated `roi.uid`).
2. Click on `Save ROI` in SlideViewer which calls `handleReportGeneration()`
3. Observe console error:

Image

   ```
   Uncaught Error: Option 'trackingIdentifier' must include a human readable tracking identifier and a tracking unique identifier.
   ```

4. `console.log(trackingIdentifier)` shows an array of three entries:
   ```json
   [
     { "uid": "…", "identifier": "ROI #1" },        ← raw options object
     { /* TextContentItem for identifier */ },
     { /* UIDRefContentItem for uid */ }
   ]
   ```

## What’s Happening

As you can see in the above image. In **templates.js** (in the [dcmjs-sr-templates](https://github.com/dcmjs-org/dcmjs/blob/master/src/sr/templates.js) ), at line **904** the length of array `TrackingIdentifier` is expected to be 2:

```
if (options.trackingIdentifier.length !== 2) {
            throw new Error(
                "Option 'trackingIdentifier' must include a human readable tracking " +
                    "identifier and a tracking unique identifier."
            );
        }
```

and the source of error lies in same **templates.js** (in the [dcmjs-sr-templates](https://github.com/dcmjs-org/dcmjs/blob/master/src/sr/templates.js) ), at line **1556** 

```
1554   class TrackingIdentifier extends Template {
1555   constructor(options) {
1556   super(options);    // <-- pushes the raw options object into `this`
1557   // …then pushes the two ContentItems…
}
```

Because `Template` extends an array, `super(options)` adds the plain `{uid, identifier}` object as the first sequence item. Downstream code expects exactly two items (TEXT + UIDREF), so it throws.

## Expected Behavior

- `TrackingIdentifier` should only produce:

  1. A `TextContentItem` (if `options.identifier` is provided)  
  2. A `UIDRefContentItem` (for `options.uid`)

- No extra plain-object entry should appear.

## Proposed Fix

In **templates.js**, change:

```diff
- constructor(options) {
-   super(options);
+ constructor(options) {
+   super();
```
after implementing this change, the issue was resolved.

nr59684:
Fix TrackingIdentifier super call

Change super(options) → super() to prevent the raw options object being injected into the sequence.

Closes #434

### DIFF
--- a/src/sr/templates.js
+++ b/src/sr/templates.js
@@ -1553,7 +1553,7 @@ class AlgorithmIdentification extends Template {
 
 class TrackingIdentifier extends Template {
     constructor(options) {
-        super(options);
+        super();
         if (options.uid === undefined) {
             throw new Error("Option 'uid' is required for TrackingIdentifier.");
         }


src/sr/templates.js
```diff
@@ -1553,7 +1553,7 @@ class AlgorithmIdentification extends Template {
 
 class TrackingIdentifier extends Template {
     constructor(options) {
-        super(options);
+        super();
```
wayfarer3130, line 1556: It isn't just this one, but every Template extension (5 of them) in this appears to be wrong.  Can you go through them and update the others?

igoroctaviano: https://github.com/dcmjs-org/dcmjs/pull/494

wayfarer3130:
This PR is being closed in favour of the general for this issue in #494

PR #443, opened 2025-07-07
feat(build): Create minified UMD build

HarmvZ:
For a project we want to use dcm.js in a browser context without build toolchain. Therefore, we would like to pull a minified version via a CDN.

This adds a minified version of the UMD build. It is ~1.10 MB vs ~1.65 MB for the original (33% reduction). 

Let me know if this looks right. I can also add a minified ESM build if that is preferred.


igoroctaviano (approved):
Looks good. Thanks!

PR #444, opened 2025-07-08
fix: Use an array of data for BufferStream

wayfarer3130:
The reader/writer for dcmjs uses a single buffer for the input and output, and this is inefficient when the buffer is coming in as smaller pieces such as when reading a DICOM file in chunks, or handling an http stream.  It also completely fails when the buffer sizes exceeds some predefined limit, typically 4 gb.

This PR modifies the BufferStream to use an array of ArrayBuffer objects for reading and writing.

The future intent of the PR is to make the groundwork to allow reading DICOM files as they are streamed, without necessarily reading the entire file into memory all at once, and to write DICOM files without keeping the entire file in memory at once, as well as to support asynchronous readers and writers.  Not all the support for this second use is available yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-node-resolve": "^15.2.3",
         "@rollup/plugin-replace": "^6.0.1",
+    "@rollup/plugin-terser": "^0.4.4",
         "acorn": "^7.1.0",
         "acorn-jsx": "^5.2.0",
         "eslint": "^9.30.1",
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -6,6 +6,7 @@ import commonjs from "@rollup/plugin-commonjs";
 // import babelRuntime from "@rollup/plugin-transform-runtime"
 import json from "@rollup/plugin-json";
 import replace from "@rollup/plugin-replace";
+import terser from "@rollup/plugin-terser";
 import { readFileSync } from "fs";
 
 const pkg = JSON.parse(
@@ -20,6 +21,13 @@ export default {
             format: "umd",
             name: "dcmjs",
             sourcemap: true
+        },
+        {
+            file: pkg.main.replace(/\.js$/, ".min.js"),
+            format: "umd",
+            name: "dcmjs",
+            sourcemap: true,
+            plugins: [terser()]
         },
         {
             file: pkg.module,
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,15 +2,28 @@
 # yarn lockfile v1
 
 
-"@ampproject/remapping@^2.2.0":
-  version "2.3.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.3.0.tgz#ed441b6fa600072520ce18b43d2c8cc8caecc7f4"
-  integrity sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==
-  dependencies:
-    "@jridgewell/gen-mapping" "^0.3.5"
-    "@jridgewell/trace-mapping" "^0.3.24"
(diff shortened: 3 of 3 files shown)


wayfarer3130:
@pieper - for this change do you think it should go in the existing codebase or should it start moving to a 2.x stream?
At the moment I'm just starting this, with the thought of taking it to the point of passing the integration tests, but not improving the basic parsing.  As a second PR, I would add different parser/write methods to allow reading using sequential blocks and reading directly to bulkdata files without necessarily storing the entire file in memory.  I would like to allow for >= 4 gb DICOM file parsing while keeping the in-memory size to something like 64 mb.

pieper:
Hi @wayfarer3130 - I don't mind the idea of a v2.0 implementation that cleans up and modernizes the buffer management, as long as we don't break backward compatibility in the process (too many things already depending on feature I think and nothing is too broken).

Regarding streaming though there's another option that I roughed out in this issue:

https://github.com/bebbi/dicom-curate/issues/8

We're currently working on using this approach for uploading large files without putting everything in memory.  I don't see the pre-PixelData tags ever exceeding 4gb, even for a WSI, even if the pixel data is larger.

Would that algorithm described there handle your use case as well?  Also, if you are doing something similar to dicom-curate maybe we should coordinate.

wayfarer3130:
@pieper - I'm trying to keep the basic interface for the streams buffer and the parser consistent, but to add other flags and a couple of other methods which can be used to have a more flexible parser, with the old versions of the parser just delegating to the new version and returning the overall final result, but allowing someone who wants more flexibility to have that by directly using the new methods.  Your suggestion of just reading the first 1 gb or something would still work in the new version, just that the new version would be able to additionally continue parsing after the 1 gb had been read if the new methods were used, and that wouldn't require re-parsing the already parsed data.   I'm modelling the parsing on the dcm4che parsing that I'm quite familiar with.

test/data.test.js
```diff
@@ -721,7 +719,7 @@ it("Writes encapsulated OB data which has an odd length with a padding byte in i
     );
 
     expect(bytesWritten).toEqual(44);
-    expect([...new Uint32Array(stream.view.buffer, 0, 11)]).toEqual([
+    expect([...new Uint32Array(stream.buffer, 0, 11)]).toEqual([
```
pieper, line 722: Is there a way to mock the `view` method so the user facing API doesn't change?

pieper:
I like the superset idea.  I added a comment about one of the tests, but looking at the other code changes maybe it would require fewer changes overall (or maybe not, I didn't look closely).  If things are all backwards compatible I don't see the need for a new version branch.

test/data.test.js
wayfarer3130: I did mock it at the next level up, just not the stream.view.buffer - I think that is the only spot it ends up being used and I'd rather not continue exposing internal details of the implementation.

wayfarer3130:
@pieper - I'm thinking that merging this PR when you are ready, and having a second PR to take advantage of it in the parser/dicom writer would be easier to review than trying to add anything more here.  It probably should be a feature version rather than a bug fix, although I was doing it primarily as a bug fix.

test/data.test.js
pieper: My only concern is that since this change shows up in a test it may also impact end-user code since the tests are basically the public API.  What will be the error if user code tries to access `view`? (undefined I guess). If we can't add a substitute for `view` can we do something that tells the downstream developer how to handle the change?

wayfarer3130: stream.view does exist and has roughly the same interface as before, down to things like slice and so on.  The changes that are different are:
1. A slice operation on the view doesn't necessarily update the original buffer - it does most of the time, but there it needs an extra call to force it to push back to the buffer.  That is all transparent when used via the stream interface.
2. The stream.buffer exists, but not the view.buffer.  The view.buffers is supposed to be accessed as stream.buffer, and AFAIK there is only the one spot it uses the wrong path - it isn't really a useful spot to access things because the getBuffer() call is the only one that gets a complete buffer that is the right size.  
3. stream.buffer isn't a mirror of hte underlying data in any way, just a copy, so it cannot be used to update.  However, the normal updates are done via stream.view and that all works.

PR #445, opened 2025-07-16
[WIP] feat: Clear the buffers as they are consumed

wayfarer3130:
This feature allows clearing buffers as they get consumed/used.  That reduces the memory footprint by allowing only as much of the object to be in memory as is required at a given time.


wayfarer3130:
Replaced by a cleaner implementation which is easier to manage against the existing API.

PR #449, opened 2025-07-30
feat: support DICOMs with missing meta length tag

ofekray:
Support missing meta length tag by reading meta headers sequentially

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -209,18 +209,61 @@ class DicomMessage {
             throw new Error("Invalid DICOM file, expected header is missing");
         }
 
+        // save position before reading first tag
+        var metaStartPos = stream.offset;
+
+        // read the first tag to check if it's the meta length tag
         var el = DicomMessage._readTag(stream, useSyntax);
+
         if (el.tag.toCleanString() !== "00020000") {
-            throw new Error(
-                "Invalid DICOM file, meta length tag is malformed or not present."
-            );
+            // meta length tag is missing
+            if (!options.ignoreErrors) {
+                throw new Error(
+                    "Invalid DICOM file, meta length tag is malformed or not present."
+                );
+            }
+
+            // read meta header elements sequentially
+            var metaHeader = {};
+            // reset stream to the position where we started reading tags
+            stream.offset = metaStartPos;
+
+            // read meta header elements until we encounter a tag outside the 0002 group
+            while (!stream.end()) {
+                var currentPos = stream.offset;
+                try {
+                    var metaEl = DicomMessage._readTag(stream, useSyntax);
+                    var tagString = metaEl.tag.toCleanString();
+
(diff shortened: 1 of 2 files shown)


ofekray:
@pieper
Thanks, I fixed them now

pieper:
@wayfarer3130 can you have a look at this?  It seems to build on https://github.com/dcmjs-org/dcmjs/commit/585e12914afd3b1000ddf737c92efc97bb3c1588

wayfarer3130:
It won't ever work with deflated transfer syntaxes, but it should be safe enough for other transfer syntaxes.  I need to read over the tests again and look at it carefully, but I think it is probably worth including.

src/DicomMessage.js
```diff
@@ -209,18 +209,61 @@ class DicomMessage {
...
+                );
+            }
+
+            // read meta header elements sequentially
+            var metaHeader = {};
+            // reset stream to the position where we started reading tags
+            stream.offset = metaStartPos;
+
+            // read meta header elements until we encounter a tag outside the 0002 group
+            while (!stream.end()) {
```
wayfarer3130, line 232: Can you use the _read method with an until tag of 00030000 ? I'd rather re-use that as much as possible

ofekray: @wayfarer3130 - Thanks, I implemented your request.

Note: The existing _read method was reading all tags if the untilTag was not found, so I added an optional "stopOnGreaterTag" parameter that causes it to stop when it encounters a tag greater than untilTag. This allows us to read only the meta header (group 0002) by using untilTag: "00030000" with stopOnGreaterTag: true.

pieper:
There was a similar feature in ITK that tried to read non-standard DICOM files with a heuristic, but I recall it needed to be disabled or modified because sometimes it would read a non-DICOM file and get an enormous number for a VL and it led to performance issues because the parser tried to read entire files when it should have failed early.

@ofekray Can you include any sanity checks to make sure this doesn't happen with this code?

ofekray:
> There was a similar feature in ITK that tried to read non-standard DICOM files with a heuristic, but I recall it needed to be disabled or modified because sometimes it would read a non-DICOM file and get an enormous number for a VL and it led to performance issues because the parser tried to read entire files when it should have failed early.
> 
> @ofekray Can you include any sanity checks to make sure this doesn't happen with this code?

Thank you for highlighting this risk. The code still checks for the DICM header at the start of parsing, so the new meta header reading logic only executes after this validation. This should prevent scenarios where a non-DICOM file is parsed. Please let me know if you think an additional sanity check is still needed, or if you have specific edge cases you would like me to validate.

pieper:
Okay, that sound reasonable then.  Since this is a known case where the DICM is present then it's not likely to have the incorrect length issue.

I'm fine with this unless @wayfarer3130 has other comments.

PR #450, opened 2025-08-06
fix: Save/load large explicit encoding values

wayfarer3130:
When encoding large explicit value encoded tags like ContourData (VR DS), dcmjs needs to write the DICOM as VR UN, according to:
   https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.8.6.html#para_23e73451-d6b6-436b-a13a-76d0e3b341b6

"If the Value Length of this Data Element exceeds 65534 bytes and an Explicit VR Transfer Syntax is used, then the Value Representation UN can be used for this Data Element. See PS3.5 Section 6.2.2."

This PR adds a unit test for this condition, with a contour data containing 64k elements (thus well over 64k for the representation of it), and adds the writing capability for this value.  The reading value for this was already handled correctly.

### DIFF
--- a/src/Tag.js
+++ b/src/Tag.js
@@ -83,19 +83,19 @@ class Tag {
     }
 
     write(stream, vrType, values, syntax, writeOptions) {
-        var vr = ValueRepresentation.createByTypeString(vrType),
-            useSyntax = DicomMessage._normalizeSyntax(syntax);
+        let vr = ValueRepresentation.createByTypeString(vrType);
+        const useSyntax = DicomMessage._normalizeSyntax(syntax);
 
-        var implicit = useSyntax == IMPLICIT_LITTLE_ENDIAN ? true : false,
-            isLittleEndian =
-                useSyntax == IMPLICIT_LITTLE_ENDIAN ||
-                useSyntax == EXPLICIT_LITTLE_ENDIAN
-                    ? true
-                    : false,
-            isEncapsulated =
-                this.isPixelDataTag() && DicomMessage.isEncapsulated(syntax);
+        const implicit = useSyntax == IMPLICIT_LITTLE_ENDIAN ? true : false;
+        const isLittleEndian =
+            useSyntax == IMPLICIT_LITTLE_ENDIAN ||
+            useSyntax == EXPLICIT_LITTLE_ENDIAN
+                ? true
+                : false;
+        const isEncapsulated =
+            this.isPixelDataTag() && DicomMessage.isEncapsulated(syntax);
 
-        var oldEndian = stream.isLittleEndian;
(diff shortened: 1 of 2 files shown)


src/Tag.js
```diff
@@ -83,19 +83,19 @@ class Tag {
...
-            isLittleEndian =
-                useSyntax == IMPLICIT_LITTLE_ENDIAN ||
-                useSyntax == EXPLICIT_LITTLE_ENDIAN
-                    ? true
-                    : false,
-            isEncapsulated =
-                this.isPixelDataTag() && DicomMessage.isEncapsulated(syntax);
-
-        var oldEndian = stream.isLittleEndian;
+        const vr = ValueRepresentation.createByTypeString(vrType);
```
wayfarer3130, line 86: Just some cleanup to this method to use the newer syntax and declare 1 value per declaration rather than using , to declare multiples.

src/Tag.js
```diff
@@ -129,12 +129,19 @@ class Tag {
...
+            valueLength >= 0x10000 &&
+            valueLength !== 0xffffffff;
+
         if (implicit) {
             stream.writeUint32(valueLength);
             written += 4;
         } else {
-            if (vr.isExplicit()) {
-                stream.writeAsciiString(vr.type);
+            if (vr.isExplicit() || isBigExplicit) {
```
wayfarer3130, line 142: And this change actually writes the UN instead.
Note the vr.isExplicit is mis-named - it SHOULD be isSeparateLength or something like that.  A vr can't be explicit, it can only have a separate length field or a combined length field.

test/data.test.js
```diff
@@ -456,6 +456,28 @@ it("test_invalid_vr_length", () => {
     }
 });
 
+it("test_long_explicit_vr", () => {
```
wayfarer3130, line 459: And here is a test for this, writing out about 200k to a single field in explicit vr.

wayfarer3130:
@pieper - I have confirmation this fixes the save issue for large RTStruct contours, and I've cleaned up a few things for the isExplicit, adding a copy of it that names it better.  It is passing the tests, and there is a test for the updated data.

pieper (approved):
Looks reasonable to me 👍 

I don't recall if this came up in the discussion but did you try the created instance with pydicom / dcmtk and dciodvfy?

wayfarer3130:
> Looks reasonable to me 👍
> 
> I don't recall if this came up in the discussion but did you try the created instance with pydicom / dcmtk and dciodvfy?

I have dciodvfy, but I will have to ask Rodrigo to test with his sample, as I dont have any samples that cause the issue to occur.

PR #453, opened 2025-08-13
fix: Use instanceof instead of constructor checks for content item, templates and value types

igoroctaviano:
- https://github.com/ImagingDataCommons/slim/issues/304


wayfarer3130 (approved):
Generally a good change - makes the code more robust.

PR #455, opened 2025-08-19
454 refactored Encoding/Decoding logic to expose DICOM encoding mappings for consumption

### LINKED ISSUE #454: Wrong decoding from ISO 2022 IR 100 to ISO IR 192 (UTF-8)?
## Intro

Hello,

I was working on a feature that I wanted to contribute to OHIF. The feature is a component for rendering encapsulated reports. During my work, I noticed that reports were showing up with `Â `, which is `Â&nbsp;`. 

I look at an internal sample DICOM SR and the report switches encoding in the `Findings` node of the SR module to `ISO 2022 IR 100`. 
Based on issues #84, #373, and #451, I am assuming that dcmjs tries to set the specific character set early on and perform conversions based on that. 
If that is the case, perhaps it is not switching to a new decoder to properly decode the input encapsulated report to `utf-8`.

I am new to dcmjs itself and it looks like this will be an involved issue. I am hoping someone with more experience can point in the right direction before I find the time to investigate the issue.

I will provide more information as I come across it.

### HTML Rendering of Report

Image

### Expected HTML Input (Different Sample at Hand in Text Editor)

Image

luissantosHCIT:
Fixes #454.

Although, I did not find any bugs with dcmjs, some of the tooling in the library could be better exposed for consumption in other projects. I focus on exposing the DICOM buffer decoding here given an initial `SpecificCharacterSet`. If translation is not necessary, a few convenience methods are provided to set the correct encoder or decoder on demand.

This is a code cleanup / reorganization kind of PR.

Jest tests passed. 
I did not add any new tests since there is coverage from the current tests and the one class I added is trivial.


src/DicomMetaDictionary.js
```diff
@@ -402,6 +419,68 @@ DicomMetaDictionary.sopClassNamesByUID = {
     "1.2.840.10008.5.1.4.1.1.77.1.5.4": "OphthalmicTomographyImage"
 };
 
+DicomMetaDictionary.encodingMapping = {
```
wayfarer3130, line 422: Can you drop this externally as it's own file?  There should also be an addEncoding method to externally add them.
just import the encoding mapping and assign to the dicom meta dictionary rather than increasing the overall contents of this file.

src/DicomMetaDictionary.js
```diff
@@ -402,6 +419,68 @@ DicomMetaDictionary.sopClassNamesByUID = {
...
+    "iso-ir-192": "utf-8",
+    gb18030: "gb18030",
+    "iso-2022-gbk": "gbk",
+    "iso-2022-58": "gb2312",
+    gbk: "gbk"
+};
+
+DicomMetaDictionary.defaultEncoding = "latin1";
+
+DicomMetaDictionary.encapsulatedSyntaxes = [
```
wayfarer3130, line 461: Same for this.  I would suggest having several categories, and use a Set for this many values.  The types:
Encapsulated
Video
Uncompressed

src/DicomMessage.js
```diff
@@ -413,18 +341,21 @@ class DicomMessage {
         } else {
             const { rawValue, value } =
                 vr.read(stream, length, syntax, options) || {};
-            if (!vr.isBinary() && singleVRs.indexOf(vr.type) == -1) {
+            if (
+                !vr.isBinary() &&
+                ValueRepresentation.singleVRs.indexOf(vr.type) === -1
```
wayfarer3130, line 346: Use ValueRepresentation.singleVRs.has(vr.type)
with a set or map.

src/DicomMetaDictionary.js
luissantosHCIT: Leaving open to remind myself to add the addEncoding method. Otherwise, I did an extensive move of these contants and converted them to Maps and Sets as requested.

src/DicomMetaDictionary.js
luissantosHCIT: Besides encapsulated, are you referring to the other syntaxes in `constants`?

src/BufferStream.js
```diff
@@ -15,20 +16,98 @@ function toFloat(val) {
     } else return val;
 }
 
+/**
+ * Facilitates the conversion of binary buffers from a DICOM encoding scheme to
+ * a web supported string encoding scheme and vice versa.
+ */
+class DicomBufferCODEC {
```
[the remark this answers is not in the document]
luissantosHCIT: Reminder for myself to add a few unit tests for this class.

src/DicomMetaDictionary.js
luissantosHCIT: There are a lot of changes in this PR so far so I will defer this item to a separate PR for now. Just pm me if there are other constants you would like me to move out of dicom.js and/or changed into Sets.

src/BufferStream.js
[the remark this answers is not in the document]
luissantosHCIT: Added two unit tests to check default and several encodings.

P.S. During this process, I found that ValueRepresentation needs some work. There are a few inconsistencies with those classes but my initial refactor (not included here) broke a lot of things and probably will need the inclusion of specific unit tests validating several classes looking to confirm:

- The expected behaviors of `writeBytes` and `applyFormatting`. 
- The number regex validation is more strict.
- Writing a `DicomDict` instance, reading the output buffer, and re-writing the generated dictionary should match the intermediate dictionary. 
- Doing test above should not yield errors with VR type generation or handling.
- Doing test above should not yield `trim` errors if a `DicomDict` instance is attempted to be written to disk.
- `write` should use the value parameter explicitly instead of using the `arguments` meta variable to get it.
- `write` should make the `ValueMultiplicity` handling less weird. It can be confusing to intuit at a glance.
- `SequenceOfItems` should write bytes directly to stream?
- Successful consolidation of duplicate logic paths.

I am going to open a separate issue and look to build a PR for those items.

For now, I am resolving this request and that should conclude the work expected in this PR's scope.

luissantosHCIT:
@wayfarer3130 Sorry for the number of commits. Every time I tried to address one aspect of this PR, I found adjacent areas that could be improved. Please, review the changes here while I split some of the refactor work to a different issue/PR.

**This PR is ready for a second review.**

src/constants/syntaxes.js
wayfarer3130, line 1: I believe encapsulated syntaxes is unused at this point - there are also more items there than are listed since new ones have been added since htis list was last updated.

src/constants/sopClassUIDs.js
wayfarer3130, line 1: This list should be generated as an artifact of the general UID list as a reverse lookup - done once it is pretty fast.

src/utilities/log.js
wayfarer3130, line 1: If you do this, you still need to export the log from the top level in the same way it was previously - there are already other users of the logging.

src/AsyncDicomReader.js
```diff
@@ -720,13 +721,13 @@ export class AsyncDicomReader {
...
             const value = vr.read(stream, length, syntax)?.value;
-            if (!vr.isBinary() && singleVRs.indexOf(vr.type) == -1) {
+            if (!vr.isBinary() && !singleVRs.has(vr.type)) {
                 values = value;
                 if (typeof value === "string") {
                     const delimiterChar = String.fromCharCode(VM_DELIMITER);
                     values = vr.dropPadByte(value.split(delimiterChar));
                 }
-            } else if (vr.type == "OW" || vr.type == "OB") {
+            } else if (vr.type === "OW" || vr.type === "OB") {
```
wayfarer3130: Should the other O types get added here?

src/BufferCODEC.js
```diff
@@ -0,0 +1,75 @@
+import { DicomMetaDictionary } from "./DicomMetaDictionary";
+import { defaultEncoding } from "./constants/encodings";
+import { selectNativeEncoding } from "./utilities/selectEncoding";
+
+/**
+ * Facilitates the conversion of binary buffers from a DICOM encoding scheme to
+ * a web supported string encoding scheme and vice versa.
+ */
+export class DicomBufferCODEC {
```
wayfarer3130: The class name and file name should match - what about:
DicomBufferCodec
Note the lower Codec here - long acronyms get treated as words with only initial upper case letter.
Should this be in utilities?

I also wonder if this should be DicomBufferText rather than Codec - the problem is that Codec generally implies image codecs in DICOM rather than text handling.

wayfarer3130: It also really has very little to do with buffers - it provides services to buffers, but it could equally provide services to other areas.  How about TextConversion or TextTranscode or DicomTextTranscode

src/BufferStream.js
```diff
@@ -453,16 +469,16 @@ export class BufferStream {
 export class ReadBufferStream extends BufferStream {
     constructor(
         buffer,
-        littleEndian,
+        littleEndian = false,
```
wayfarer3130: Big Endian encoding in DICOM is deprecated so probably littleEndian=true should be the default here if nothing provided.  Ideally I'd like to move this option to the options, but that is a (small) breaking API change.  @pieper - what do you think - should we document an API change in favour of being more consistent with DICOM for the standard setup of the buffers?  This is MOSTLY an internal detail so I would be in favour of doing it.

src/BufferStream.js
```diff
@@ -564,8 +576,8 @@ export class DeflatedReadBufferStream extends ReadBufferStream {
 }
 
 export class WriteBufferStream extends BufferStream {
-    constructor(defaultSize, littleEndian) {
-        super({ defaultSize, littleEndian });
+    constructor(defaultSize, littleEndian, encoding = defaultDICOMEncoding) {
```
wayfarer3130: The encoding is incorrect to pass at the beginning like this, and definitely should get added as additional options object if you do want it.  The reason is the way the DICOM is written into a buffer is that the initial encoding is ASCII until the specific character set gets written, and it then changes to the specific character set encoding.  

There are also other options it would be nice to include here for streaming output, and those are better passed in an optional object.

Deleting littleEndian and assuming it instead is also a good idea, as the only time one starts with a big endian buffer is in writing DIMSE output where the FMI data is in a separate block from the DICOM encoded data.  Even for that case, it starts in ASCII (7 bit data) and only switches once the specific character set is written.

src/constants/encodings.js
```diff
@@ -0,0 +1,39 @@
...
+    ["iso-2022-ir-166", "tis-620"],
+    ["iso-2022-ir-58", "iso-ir-58"],
+    ["iso-ir-192", "utf-8"],
+    ["gb18030", "gb18030"],
+    ["iso-2022-gbk", "gbk"],
+    ["iso-2022-58", "gb2312"],
+    ["gbk", "gbk"]
+]);
+
+export const defaultEncoding = "latin1";
```
wayfarer3130: Technically, the default encoding is ascii, not latin1 - at a minimum a comment indicating that is required, but assuming latin1 can get you into troubles with incorrect character data.

src/constants/encodings.js
```diff
@@ -0,0 +1,39 @@
+export const encodingMapping = new Map([
+    ["", "iso-8859-1"],
+    ["iso-ir-6", "iso-8859-1"],
+    ["iso-ir-13", "shift-jis"],
+    ["iso-ir-100", "latin1"],
```
wayfarer3130: Please use iso-8859-1 instead of latin1 for consistency.

src/DicomTextTranscode.js
wayfarer3130, line 1: If you are going to make this change, could you add decode/encode support for patient name multi-character set encodings and let the class handle that?

src/DicomDict.js
```diff
@@ -22,13 +23,28 @@ class DicomDict {
         }
     }
 
-    write(writeOptions = { allowInvalidVRLength: false }) {
-        var metaSyntax = EXPLICIT_LITTLE_ENDIAN;
-        var fileStream = new WriteBufferStream(4096, true);
+    write(
```
wayfarer3130, line 25: These options don't belong here - they happen as a result of the internal encoding changes.  
Ideally the constructor should be new WriteBufferStream() or new WriteBufferStream(options) to allow passing undeclared options through.

src/DicomDict.js
```diff
@@ -22,13 +23,28 @@ class DicomDict {
...
+            4096,
+            writeOptions.littleEndian,
+            writeOptions.encoding
+        );
+
         fileStream.writeUint8Repeat(0, 128);
         fileStream.writeAsciiString("DICM");
 
-        var metaStream = new WriteBufferStream(1024);
+        const metaStream = new WriteBufferStream(
```
wayfarer3130: This is completely wrong here - the metaStream does NOT using hte encoding at all, and is always little endian.  Again, 
new WriteBufferStream()  is a better choice not passing any arguments, just getting the defaults.

src/DicomMessage.js
```diff
@@ -75,33 +75,19 @@ export class DicomMessage {
...
-                                readInfo.values
-                            );
-                        } else {
-                            throw Error(
-                                `Using multiple character sets is not supported: ${readInfo.values}`
-                            );
-                        }
-                    }
-                    readInfo.values = ["ISO_IR 192"]; // change SpecificCharacterSet to UTF-8
+                    // Note, I have seen a switch of encoding when entering SR nodes.
```
wayfarer3130, line 78: That is an illegal change and should be ignored - it does not refer to the encoding of the rest of the DICOM data underneath it, but refers to the encoding of a referenced object.  Both reading and writing should not update the encoding except when specific character set is at the top level.

src/DicomMessage.js
```diff
@@ -150,23 +136,24 @@ export class DicomMessage {
             forceStoreRaw: false
         }
     ) {
-        var stream = new ReadBufferStream(buffer, null, {
-                noCopy: options.noCopy
-            }),
-            useSyntax = EXPLICIT_LITTLE_ENDIAN;
+        let stream = new ReadBufferStream(buffer, false, {
```
wayfarer3130: The null was very ugly previously as it wasn't clear what was getting applied there.

src/BufferCODEC.js
luissantosHCIT: > DicomBufferCodec

That's a good point. I meant CODEC more as any algorithm that codifies and decodifies, but **you are right others might become confused**...

Names, are so hard. Haha!

I'll settle for `DicomTextTranscode`. I think that suggestion works closest for my brain too.

src/BufferStream.js
luissantosHCIT: good point.

src/utilities/log.js
luissantosHCIT: I'll do a file search to check where that export should be remapped/live.

src/AsyncDicomReader.js
luissantosHCIT: how about `vr.isBinary()` like I noticed the other day? I think the idea here is to return the data if it is a binary blob, right?

src/DicomMessage.js
luissantosHCIT: Well, let me tell you about a little set of companies named `Philips` and `CareStream` that do switch encoding within the `Findings SR` node probably because the generator of the HTML version of a report was targetting UTF-8 instead of the default ISO 8859_1 of the rest of the header. Not accommodating the switch (which was what triggered this PR in the first place) means my PR on the OHIF side would have to remain with a hack to ensure that the string content is converted to UTF-8 in that case. However, what if others have taken the same liberty to switch to a different character set that is not UTF-8? 

In my opinion, the long term solution would be to temporarily switch character set upon walking through an SR node if it requests to do so. That probably means that parsing of the DICOM Header would have to be broken into a non-SR pass and an SR-pass. I am not in a position to make that bigger change right now. **The logic that was already there before my changes was set up to do the switch with the expectation that the character set tag would hopefully only happen once.** I think that is in itself a separate bug/issue. I simply cleaned the logic a bit to allow its usage elsewhere (hence `selectEncoding`).

I also need to see if there is any prescription by the DICOM standard on whether the encoding should be handled once or on demand. I think the SR paradigm introduces the concept of `scope` implicitly and I am not sure people thought about how it clashes with the one dimensional approach of the header (at least not based on anecdotes from someone who knew one of the people who worked on the SR module). Unfortunately, right now, if you want to be able to read the Philips PACS reports, you have no choice but to switch encoding.

src/DicomMessage.js
luissantosHCIT: Yeah, this is related to your suggestion to simply use littleEndian implicitly and set to true by default. I agree with that approach unless someone out there has been trying to do DIMSE in JavaScript for some weird reason (`dcmtk` exists for that).

src/DicomTextTranscode.js
luissantosHCIT: I believe you mean the specialization of `VR`. **I will see what I can do here**. 

 I am already working on a more thorough pass through VR which has uncovered some interesting behaviors. I would rather tackle long term design decisions there (#483).

src/BufferStream.js
pieper: Yes, I think defaulting to little endian makes sense.

src/DicomMessage.js
pieper: We'd like to keep the core of dcmjs clear of any special cases to handle incorrect dicom.  The pattern that we have already is the concept of "normalizers" that are partial specialization subclasses that detect a particular variant and convert the dataset into a form that you would expect according to the standard.  That way we leave all the hacks located in one place and don't accidentally apply them in the wrong place.  I've usually considered that pattern to apply after the data is read, but perhaps it is needed here too.

src/DicomMessage.js
pieper: There is in fact a dicom-dimse package : )

https://github.com/PantelisGeorgiadis/dcmjs-dimse

luissantosHCIT: Incredible. 

@pieper, I am guessing we still want to default to little endian. Would you prefer we keep the constructor signatures defaulting to true (no removal of the littleEndian parameter)? Simplify the constructor to get closer to `ReadBufferStream()`?

pieper: I don't have a strong opinion on this issue specifically, but simpler and more consistent sounds nice to me.

src/DicomMessage.js
luissantosHCIT: @pieper, I agree with such a pattern. I have used it myself for tackling the normalization of SR trees in a Python pipeline framework I developed. 

Would it be possible to defer this larger refactor of how this code path handles the encoding to a separate issue? I feel the scope keeps getting bigger for this particular PR. That's why I separated out improvements to VR into its own issue and PR. I want to focus on wrapping up what is currently present here at this moment.

src/constants/encodings.js
luissantosHCIT: I will make it `ascii`.

src/DicomMessage.js
pieper: Yes, keeping things focused is important.  Since several groups use this in production I'm very hesitant to make changes.  So lots of testing is key.

src/DicomMessage.js
wayfarer3130: Then lets make a small breaking change, document it in a migration guide for the small number of cases that people are creating buffers externally (that is mostly me in static dicomweb and the dicom-dimse), and then simplify the API to take an options object, dont throw if it passes in a number for the size and use the simple constructor version by default.

src/utilities/log.js
luissantosHCIT: As far as I could see, `index.js` import of `log` was remapped properly to utilities and the other source files are pointing to the correct location of the log source. If there is another area I am missing that comes to mind let me know.

src/AsyncDicomReader.js
luissantosHCIT: @wayfarer3130 , hmm I tried changing to `isBinary()` so we could inclusively return the value as is if the vr is a binary vr, but that breaks tests (more than one). For now, I am not going to poke this harder since I am short on my time budget. I have a feeling that I will need to revisit this issue and thoroughly review the tests to make sure I respect the spirit of those tests.

luissantosHCIT: Ignore above. I went back to https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html and added the other O VRs as explicit checks like found here and reran the tests successfully. I have a commit to address this request. :)

src/constants/sopClassUIDs.js
luissantosHCIT: I am not sure what you mean in this case, so I am skipping for now. Feel free to elaborate and if the PR is still open, I might try making it happen. :)

src/DicomMessage.js
luissantosHCIT: Ok, I made the changes discussed. I need to look into the documentation side of things tomorrow.

luissantosHCIT: I added a documentation note under docs.

luissantosHCIT:
@pieper and @wayfarer3130, I made the last set of changes requested. While remaining within scope. I kept the changes in this pass strictly to what was requested. I went ahead and opened issues #480 (attempting to address items that I found during VR handling) and #484 (to remind me to refactor the handling of character set switching per our discussion). Let me know if the current changes work for everyone. 

Thank you for your time here!

pieper:
Looks like a couple remaining check issues.

I'm thinking someone should try building OHIF or other projects off this branch for some testing.  I wonder if we should add that as a build check on PRs to get a netlify hosted test version or similar.

luissantosHCIT:
> Looks like a couple remaining check issues.
> 
> I'm thinking someone should try building OHIF or other projects off this branch for some testing. I wonder if we should add that as a build check on PRs to get a netlify hosted test version or similar.

I was going to do that after the merge, but you are right. 

I will do that now instead in my OHIF PR branch to see if there are issues I will need to address.

wayfarer3130:
> > Looks like a couple remaining check issues.
> > I'm thinking someone should try building OHIF or other projects off this branch for some testing. I wonder if we should add that as a build check on PRs to get a netlify hosted test version or similar.
> 
> I was going to do that after the merge, but you are right.
> 
> I will do that now instead in my OHIF PR branch to see if there are issues I will need to address.

There is a change just being worked on right now in CS3D that allows building OHIF previews/running tests and running the OHIF PR tests with a just built CS3D version - it would be very nice to have the same sort of thing.

The OHIF side works well now, but the CS3D side isn't quite there yet.

luissantosHCIT:
A few updates from my side:

- I cleared up the reported unused imports.
- I found I was using a much newer version of node and that is why I was not getting errors when running the test suites, but I was able to replicate the errors now. I am addressing the errors now.
- Modifying the package.json to include  `"dcmjs": "https://github.com/luissantosHCIT/dcmjs.git#454_incorrect_decoding_of_iso_ir_100"` in every workspace generates `module not found` errors after running `yarn install`. Do I need to do something else to temporarily switch to use this branch to test the OHIF Viewer? @wayfarer3130

luissantosHCIT:
Assuming I did not miss an unused import, all that is left is for me to figure out how to make OHIF accept my branch in package.json.

luissantosHCIT:
Nvm, I figured it out. I needed to import from my local project directory instead from github because the repo lacks a copy of the built bundle by default. Oops.

I am currently testing the viewer using my built copy of dcmjs.

luissantosHCIT:
Tests done with my OHIF branch pointing to this dcmjs branch:

- [x]  All unit tests
- [x]  Cypress E2E tests
- [x]  Manual opening of cases 
- [x]  Manual navigation
- [x] Manual review of one of the Cypress tests

The Cypress tests showed an error. This error was present regardless of which version of `dcmjs` was in use. The error did not occur when I manually navigated to the case and opened it following the test's steps. 

Screenshot From 2026-03-18 13-02-57

**My conclusion is that the dcmjs changes here do not affect the OHIF side. Do try yourselves in case I missed a way to force bugs out of this dcmjs repo from the OHIF Viewer.**

@pieper @wayfarer3130

fedorov:
@igoroctaviano do you want to take a look at this - mostly to make sure this does not negatively affect functionality important for IDC?

igoroctaviano:
> @igoroctaviano do you want to take a look at this - mostly to make sure this does not negatively affect functionality important for IDC?

From static analysis alone of the code, for dicom-microscopy-viewer, ohif fork and slim looks like this does not introduce compatibility problems. Residual risk is behavioral, like string decoding for certain character sets, but I ran slim/dmv/ohif linked here, and it seems fine..

luissantosHCIT:
@pieper Just a small poke to ask if there is anything else needed to merge this change. 
Thank you!

pieper:
Sorry for the delays @luissantosHCIT - I've become a bit gunshy in merging.  This one's a lot to parse, especially with all the formatting and other small changes that make it hard to see the big picture.  I kind of wish those were a separate PR but that may just be extra work at this point.

Looks like it's now got a conflict with the main branch.

Assuming that gets resolved it sounds like Igor is okay so maybe @wayfarer3130 can chime in if all his issues got resolved.

luissantosHCIT:
> Sorry for the delays @luissantosHCIT - I've become a bit gunshy in merging. This one's a lot to parse, especially with all the formatting and other small changes that make it hard to see the big picture. I kind of wish those were a separate PR but that may just be extra work at this point.
> 
> Looks like it's now got a conflict with the main branch.
> 
> Assuming that gets resolved it sounds like Igor is okay so maybe @wayfarer3130 can chime in if all his issues got resolved.

I completely understand. This is a foundational project that affects many people!

Let me see what the conflict is and resolve it for you.

I am just making sure that the PR does not get forgotten if it can be merged since at some point I might just forget about it as well. Haha.

luissantosHCIT:
Looks like in this PR I had moved the VR constants to their own location for usage in multiple sources but main had the pre-PR change. I kept my change here since we handle the refactor here.

This reminds me of the VR refactor/clean up I was trying to do a while back and which I forked into its own ticket. I am currently working on optimizing my Rust-based HL7 interface toolkit so I will not get to finish that work (#480) for some time now. If it sits without progress for a year, feel free to ask me to close it or you can close it yourself at that point.

I agree about keeping the PRs simpler moving forward. I prefer to reach "checkpoints" in a complex objective myself. I will try to improve on where to draw the line on changes and break the PRs further.

PR #463, opened 2025-12-15
revert: Changes from streaming handler

wayfarer3130:
There were issues found when running the streaming data parser, and using the async handling in JavaScript seems to produce an easier to understand version, so this change is being reverted in favour of the easier to understand version.

A limited number of changes are kept for things like constants and some jest logging configuration.


wayfarer3130:
@pieper - I will merge this tomorrow if you don't have any feedback.  It is basically just:
git revert SHA1SUM
on the streaming parser.

pieper:
That sounds good to me, thanks.

pieper (approved):

PR #464, opened 2025-12-15
fix: async parse class

wayfarer3130:
Add a unit test for reading partial buffers and fix the related partial buffer read operations.

Add a streaming input read based on the partial buffer input stream.

### DIFF
--- a/src/AsyncDicomReader.js
+++ b/src/AsyncDicomReader.js
@@ -7,7 +7,8 @@ import {
     VM_DELIMITER,
     UNDEFINED_LENGTH,
     TagHex,
-    encodingMapping
+    encodingMapping,
+    UNDEFINED_LENGTH_FIX
 } from "./constants/dicom";
 import { Tag } from "./Tag";
 import { DicomMessage, singleVRs } from "./DicomMessage";
@@ -60,8 +61,7 @@ export class AsyncDicomReader {
         const listener = options?.listener || new DicomMetadataListener();
         this.dict ||= {};
         listener.startObject(this.dict);
-        await this.read(listener, options);
-        this.dict = listener.pop();
+        this.dict = await this.read(listener, options);
 
         return this;
     }
@@ -118,10 +118,11 @@ export class AsyncDicomReader {
     }
 
     async read(listener, options) {
+        const untilOffset = options?.untilOffset || Number.MAX_SAFE_INTEGER;
         this.listener = listener;
         const { stream } = this;
         await stream.ensureAvailable();
-        while (stream.isAvailable(1, false)) {
+        while (stream.offset < untilOffset && stream.isAvailable(1, false)) {
             // Consume before reading the tag so that data before the
             // current tag can be cleared.
             stream.consume();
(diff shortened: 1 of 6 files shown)


src/BufferStream.js
wayfarer3130: This does change the BufferStream for the current DicomMessage class, but I also added a unit tests for this so as to make sure that the new functionality works as expected

src/AsyncDicomReader.js
```diff
@@ -0,0 +1,412 @@
...
+
+        return this;
+    }
+
+    async readFmi(options = null) {
+        const { stream } = this;
+        await stream.ensureAvailable();
+        const { offset: metaStartPos } = stream;
+        const el = this.readTagHeader();
+        if (el.tag !== "00020000") {
```
pieper: I prefer never to have raw hex in the code.  Can you use a symbol from the dictionary here? (and everywhere)

src/AsyncDicomReader.js
```diff
@@ -0,0 +1,412 @@
...
+        const listener = options?.listener || new DicomMetadataListener();
+        this.dict ||= {};
+        listener.startObject(this.dict);
+        await this.read(listener, options);
+        this.dict = listener.pop();
+
+        return this;
+    }
+
+    async readFmi(options = null) {
```
pieper: Not clear what Fmi is here.  Can you use a more descriptive symbol to make the code more self-documenting?

src/AsyncDicomReader.js
```diff
@@ -0,0 +1,412 @@
...
+            if (elementData) {
+                vrType = elementData.vr;
+            } else {
+                //unknown tag
+                if (length == 0xffffffff) {
+                    vrType = "SQ";
+                } else if (tagObj.isPixelDataTag()) {
+                    vrType = "OW";
+                } else if (vrType == "xs") {
+                    vrType = "US";
```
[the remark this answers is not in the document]
wayfarer3130: It is tricky to handle generically since it isn't always consistent what it depends on.  USUALLY it is the signed status of the data, but especially for streaming handling, the signed state isn't necessarily available in time.

src/AsyncDicomReader.js
wayfarer3130: Fixed - I didn't use the dictionary here since I'd like the code to mostly be agnostic about needing a full dictionary.

src/DicomDict.js
wayfarer3130: I also use the constant in the DicomDict and DicomMessage

wayfarer3130:
@pieper - I updated the PR addressing your comments, except the "ox" one, which I don't have a good resolution for until that other PR is merged.

wayfarer3130:
@pieper - added the xs conversion if we have PixelRepresentation, SpecificCharacterSet decoding and a couple more tags.

src/AsyncDicomReader.js
```diff
@@ -0,0 +1,447 @@
...
+            length = stream.readUint32();
+            vr = ValueRepresentation.createByTypeString("UN");
+        } else if (implicit && !isCommand) {
+            length = stream.readUint32();
+            var elementData = DicomMessage.lookupTag(tagObj);
+            if (elementData) {
+                vrType = elementData.vr;
+            } else {
+                //unknown tag
+                if (length == 0xffffffff) {
```
pieper: Lingering hex

pieper (changes requested):
Just a few tweaks.

wayfarer3130:
@pieper - made the two remaining fixes.
Also fixed fragmented data reading.  It isn't fully streaming within the image, but between images is still fully streaming.  The reason is that until you have read the next item, it is difficult to figure out the status of the last one.  Probably I will try fixing that in the future, but I think it is ok for now.

PR #470, opened 2026-01-28
feat: Add a packed/fast loading dictionary.js 

wayfarer3130:
This change compresses the dictionary quite a lot, having far fewer in-memory objects required for loading the dcmjs side.  Together with the async loader it improves quite a lot on the parsing of DICOM as compared to dicom-parser and is recommended as a complete replacement for that library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,56 @@ git-cz --non-interactive --type=fix --subject="commit message"
 
 More info at [git-cz](https://www.npmjs.com/package/git-cz).
 
+## DICOM Dictionary
+
+The dcmjs library includes DICOM data dictionaries that map DICOM tags to their metadata (VR, VM, etc.). To optimize load performance, the library uses a pre-compiled "fast dictionary" format.
+
+### Dictionary Files
+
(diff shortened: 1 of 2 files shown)


wayfarer3130:
@sedghi - performance metrics added, here is a sample run.  On average the newer version loads about 5 times faster for just the dictionary portion, about a 100ms savings - I swapped the order to test it, and order does make a difference, but not much of one.

--- Dictionary load benchmark (Bun) ---

Old dictionary (generate/dictionary.mjs): 181.16 ms
New dictionary (src/dictionary.fast.js):  19.04 ms
Old/New: 9.52x (new faster)

ESM main (dcmjs.es.js):        112.01 ms
ESM private (loadPrivateTags): 0.01 ms
ESM total:                    112.01 ms

UMD (dcmjs.js):               72.11 ms


ESM loads more bits so it takes a bit longer - that will not be true in production usage with webpack builds.

wayfarer3130:
Size metrics:
dictionary.js 1,350 kb
194,294 dicom.packed.js
177,242 dictionary.private.data.js
11,408 dictionary.ranges.data.js

About 400 kb packed versus 1,350 original.

Note the original is still available as an esm import directly.

PR #472, opened 2026-01-31
fix: unencapsulatedTransferSyntaxes mapping

Michael-Andersen:
I noticed this mapping wasn't set up correctly which was causing unencapsulated pixel data to be written as encapsulated.

### DIFF
--- a/src/Tag.js
+++ b/src/Tag.js
@@ -136,12 +136,10 @@ class Tag {
         const vr = ValueRepresentation.createByTypeString(vrType);
         const useSyntax = DicomMessage._normalizeSyntax(syntax);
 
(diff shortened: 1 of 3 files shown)


wayfarer3130:
@Michael-Andersen - would it be possible for you to add a test for the unencapsulated mapping?  It would be nice to know we aren't writing that incorrectly in the future.  You can use the stop before tag parsing to get up to the point of pixel data, then use the DicomMessage parsing directly to get at the internal format of the generated data.
If you ask AI to generate a test that does that for you and creates static static for a test encoding, it generally works pretty well.

wayfarer3130 (approved):

PR #475, opened 2026-02-26
fix: Update to newer loglevel

wayfarer3130:
This change updates dcmjs to use loglevel 1.9.2 which adds improved ability to set the log levels hierarchically.
It also uses the global loglevel instance if present to allow sharing a single loglevel instance, which allows redirecting logs effectively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,3 +222,5 @@ The developers gratefully acknowledge their research support:
 ## Logging
 
 This library uses [loglevel](https://github.com/pimterry/loglevel) for logging. By default, the log level is set to "warn". You can change the log level by setting the `LOG_LEVEL` environment variable or by using the `setLevel` method in your code.
+
+Named loggers (for example `validationLog`) can be turned up or down independently of the root log and of each other. That lets you enable validation messages without changing general logging, or silence validation in tests while keeping other logs. Use `getLogger(name)` for custom loggers, or the built-in `validationLog` (see `src/log.js`). Set level with e.g. `validationLog.setLevel("warn")` or `validationLog.setLevel(5)` for silent. Examples in this repo: validation truncation is logged at error level in [`src/DicomMetaDictionary.js`](src/DicomMetaDictionary.js) (search for `validationLog.error`), and a test that suppresses that output by setting the validation logger to silent only for that test is in [`test/data.test.js`](test/data.test.js) (search for `test_code_string_vr_truncated`).
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,40 +1,20 @@
-const { Console } = require("console");
-const nodeConsole = new Console(process.stdout, process.stderr);
+// Jest setup - uses real loglevel, writing to stdout to avoid stack traces from console.warn/error.
+// Format: loggerName [LEVEL] message
+// Log level is controlled by LOG_LEVEL env (default "warn" in src/log.js).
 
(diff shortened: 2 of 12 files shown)


wayfarer3130:
@pieper - this is just an update to a newer loglevel with a couple of log reductions for things that happen commonly like the ox logs - those are handled externally in the AsyncDicomReader version so can have a debug level log.

pieper (changes requested):
Is this backwards compatible?

Can we have a test demonstrating how to use any new features this enables?

wayfarer3130:
> Is this backwards compatible?
> 
> Can we have a test demonstrating how to use any new features this enables?

Yes, it is backwards compatible, just quite a bit more flexible.  I've updated the PR to give examples in the README about how to set the loglevels on different loggers, with examples from the jest unit tests, now using the real loglevel code rather than a mock.

pieper:
I asked some dcmjs users to comment but one guy is on vacation this week.  Let's hold off and see if we get more comments.

bebbi:
@wayfarer3130 I took a look at this PR and the loglevel upgrade + log-level downgrades look good. I created a branch in my fork and put two commits on top of yours to avoid the duck-type check (`typeof globalThis.log.getLogger === "function"`), which is a bit loose and could match non-loglevel objects like winston or pino that happen to have a `getLogger` method on a generic global name.

https://github.com/bebbi/dcmjs/tree/update/loglevel-version

The idea is:
1. Replace the `globalThis.log` sniffing with an explicit `setRootLogger(root)` function that consumers call to share their loglevel instance. Loggers are lazily initialized so the root can be swapped at any time.
2. Switch all internal call sites to use getter functions (`getLog().warn(...)` instead of `log.warn(...)`) so there's no Proxy indirection in the hot path.
3. Keep the old `log` and `validationLog` exports around as deprecated Proxy wrappers that emit a one-time `console.warn`, so existing consumers don't break immediately. Removing them later is a ~20-line deletion across two files.

What do you think?

bebbi:
@wayfarer3130 I only noticed the force-push now and the code looks a bit different now - do the 2 commits on top of your previous state clarify the intent?

wayfarer3130:
> @wayfarer3130 I only noticed the force-push now and the code looks a bit different now - do the 2 commits on top of your previous state clarify the intent?

I had realized I had started this a second time, only the second time was working more consistently with the provided code to set the shared logger, so I force pushed to remove the old code version.

I don't think your suggested setRootLogger stuff is worthwhile - the loglevel code already has ways to replace the output end, as well as globally set the log levels starting with loglevel.getLogger on the full root path.  It also shares a global loglevel instance, and it can be replaced in the dcmjs code by a resolution at the top level module, so I think we don't need the extra complexity.

bebbi:
The intention of my commits was a long-term cleanup that removes the duck typing of the global logger object. I added the proxy as a temporary measure to enable smooth deprecation, it should be removed again after a grace period.
I’ll leave the decision to you though.

wayfarer3130:
> The intention of my commits was a long-term cleanup that removes the duck typing of the global logger object. I added the proxy as a temporary measure to enable smooth deprecation, it should be removed again after a grace period. I’ll leave the decision to you though.

For the logger object, I don't see any reason not to allow the duck typing of it - the point is to allow a replacement to exist, so duck typing seems appropriate.

Issue #480, opened 2026-03-10
Refactor of ValueRepresentation.


luissantosHCIT:
During my work in #455, I found that `ValueRepresentation` needs some work. There are a few inconsistencies with those classes but my initial refactor (not included here) broke a lot of things and probably will need the inclusion of specific unit tests validating several classes looking to confirm:

- [ ] The expected behaviors of `writeBytes` and `applyFormatting`. 
- [ ]  The number regex validation is more strict.
- [ ]  Writing a `DicomDict` instance, reading the output buffer, and re-writing the generated dictionary should match the intermediate dictionary. 
- [ ] Doing test above should not yield errors with VR type generation or handling.
- [ ] Doing test above should not yield `trim` errors if a `DicomDict` with `_rawValue` instance is attempted to be written to disk.
- [ ]  `write` should use the value parameter explicitly instead of using the `arguments` meta variable to get it.
- [ ] `write` should make the `ValueMultiplicity` handling less weird. It can be confusing to intuit at a glance.
- [ ]  `SequenceOfItems` should write bytes directly to stream?
- [ ]  Successful consolidation of duplicate logic paths.

*Note:* Work here might address precision loss discussed in #479, since I have been playing with a different regex approach to filter/validating numbers. I would certainly be happy to build a unit test after I publish my refactor to check for that issue.

PR #483, opened 2026-03-11
[WIP] 480 refactor vr

luissantosHCIT:
See #480.

This is my attempt at addressing issues I identified with `ValueRepresentation` as well as code cleanup.
I am also creating a `test harness` here to ensure key methods and `ValueRepresentation types` maintain the expected behavior pre and post refactor. Perhaps, those test will help others confidently refactor `VR` further in the future.


luissantosHCIT:
As I update the unit test harness, I realized that to test the `readEncodedString` family of methods #455 will need to be merged in or the new tests will fail in the future due to the broken encoding/decoding before that PR. I do not directly invoke this method but I made a refactor to use the `padded` version in the `ValueRepresentation` class which handles the pad byte more gracefully.

PR #488, opened 2026-04-06
feat: Support for VR type "UV" 

PolarsBear:
In accordance with the (not so) new DICOM metadata standards, since 2019a, as stated in PS3.5, section 6.2 (see [here](https://dicom.nema.org/medical/dicom/2019a/output/chtml/part05/sect_6.2.html#para_570df2d2-ab6d-4bfa-b6cb-cad800e4596c)), the VR type "UV", or "Unsigned 64 bit Very Long" describes, as one would expect, a 64 bit unsigned integer.

To comply with these standards and avoid interpreting these values as VR type "UN" (Unknown), I offer this contribution.

I attempted to keep things as close to the standard you've set as I could, but, in case of any insatisfaction, I am more than capable of reworking the code herein suggested.

The tests leave quite a bit to be desired. I must admit testing is not my greatest strong suit. Basic testing is in place for reading and writing, but more in-depth tests with real DICOM datasets would be welcome.

I couldn't find any documentation as to how I should write the Pull Request, so I hope this is acceptable

(Also, my first Pull Request, so I ask you to forgive me if something here isn't up to standard)

### DIFF
--- a/src/ValueRepresentation.js
+++ b/src/ValueRepresentation.js
@@ -66,7 +66,7 @@ function toWindows(inputArray, size) {
 let DicomMessage, Tag, DicomMetaDictionary;
 
 var binaryVRs = ["FL", "FD", "SL", "SS", "UL", "US", "AT"],
-    length32VRs = ["OB", "OW", "OF", "SQ", "UC", "UR", "UT", "UN", "OD"],
+    length32VRs = ["OB", "OW", "OF", "SQ", "UC", "UR", "UT", "UN", "OD", "UV"],
     singleVRs = ["SQ", "OF", "OW", "OB", "UN"];
 
 class ValueRepresentation {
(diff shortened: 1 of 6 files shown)


pieper (approved):
This seems reasonable to me, thanks for the contribution 👍 

I'd like another set of eyes though, so maybe @wayfarer3130 can take a look.

src/ValueRepresentation.js
```diff
@@ -1359,6 +1359,29 @@ class UnsignedLong extends ValueRepresentation {
...
+    constructor() {
+        super("UV");
+        this.maxLength = 8;
+        this.padByte = PADDING_NULL;
+        this.fixed = true;
+        this.defaultValue = 0;
+    }
+
+    readBytes(stream) {
+        return stream.readBigUint64();
```
wayfarer3130, line 1372: This code is incorrect on two counts - it doesn't allow repeated UV values, and it isn't returning an array.  Just pushing a change now to fix that.

wayfarer3130: Actually, the binary vrs will re-run the read/write multiple times, so I've just added UV to the binary VRs list.

wayfarer3130 (approved):
Fixed the issue with the encoding type and added some equivalency tests.

Issue #478, opened 2026-03-02
AsyncDicomReader().read({ untilTag: TAG, includeUntilTag: false }) is broken


Iman974:
Hello again dear devs,

I encountered 2 problems with `AsyncDicomReader().read()`.

### 1. AsyncDicomReader().read() crashes with includeUntilTag=true

[This piece of code](https://github.com/dcmjs-org/dcmjs/blob/master/docs/AsyncDicomReader-skill.md#reading-until-specific-tag) in the AsyncDicomReader Skill Guide crashes. After reading `AsyncDicomReader().readFile()`, I figured out that it is because you forgot to mention we need to use `listener.startObject()` before calling `read()`, and not call `listener.pop()` again after getting the result:
```javascript
async function readUntilTag(arrayBuffer, targetTag) {
  const reader = new AsyncDicomReader();
  reader.stream.setData(arrayBuffer);
  
  await reader.readPreamble();
  await reader.readMeta();
  
  const listener = new DicomMetadataListener();
  listener.startObject({}) // => This was missing and causes an error when calling the first pop()

  const result = await reader.read(listener, {
    untilTag: targetTag,
    includeUntilTagValue: false
  });
  
  return result; // Now don't call listener.pop() again or it will throw an error
}
```
Here is the error with the initial piece of code:
```javascript
Uncaught (in promise) TypeError: can't access property "pop", this.current is null
    _basePop DicomMetadataListener.js:270
    _callee5$ AsyncDicomReader.js:301
    tryCatch dcmjs.es.js:584
    makeInvokeMethod dcmjs.es.js:672
    defineIteratorMethods dcmjs.es.js:613
    asyncGeneratorStep dcmjs.es.js:868
    _next dcmjs.es.js:887
    promise callback*asyncGeneratorStep dcmjs.es.js:877
    _next dcmjs.es.js:887
    // ...
    // This goes on many times
    // ...
    promise callback*asyncGeneratorStep dcmjs.es.js:877
    _next dcmjs.es.js:887
DicomMetadataListener.js:270:21
```

So I would suggest either adding this to the documentation, or if we actually always want a listener to have a non-null start object, then perhaps modifying `DicomMetadataListener`'s constructor to always create a start object would be ideal, although this may require removing some calls to `listener.startObject({})` in a few places of `AsyncDicomReader`.

[...]

PR #490, opened 2026-04-13
fix: avoid reading data once untilTag is encountered

LubosD:
In the async reader, when readTagHeader() returns {untilTag: true}, the loop doesn't break - it falls through to listener.addTag / readPixelData with a garbage length, corrupting stream.offset (that I intend to use to handle given tag contents on my own).

May be related to #478

@pieper 🙏

### DIFF
--- a/src/AsyncDicomReader.js
+++ b/src/AsyncDicomReader.js
@@ -267,6 +267,16 @@ export class AsyncDicomReader {
             // current tag can be cleared.
             stream.consume();
             const tagInfo = this.readTagHeader(options);
+
+            // Stop when the requested tag boundary is reached.  readTagHeader()
+            // has already consumed the 4-byte tag but nothing beyond it, so
+            // stream.offset now points at the first byte after the tag (i.e.
+            // the VR field for explicit-LE).  Callers that need the start
+            // offset of the tag itself should subtract 4 from stream.offset.
+            if (tagInfo.isUntilTag) {
+                break;
+            }
+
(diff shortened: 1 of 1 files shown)


pieper (approved):
This sounds right to me.  Can you weigh in @wayfarer3130 ?

LubosD:
Hey @wayfarer3130 & @pieper , can we merge this? :-) It's a really simple fix after all.

src/AsyncDicomReader.js
```diff
@@ -267,6 +267,16 @@ export class AsyncDicomReader {
             // current tag can be cleared.
             stream.consume();
             const tagInfo = this.readTagHeader(options);
+
+            // Stop when the requested tag boundary is reached.  readTagHeader()
+            // has already consumed the 4-byte tag but nothing beyond it, so
+            // stream.offset now points at the first byte after the tag (i.e.
+            // the VR field for explicit-LE).  Callers that need the start
+            // offset of the tag itself should subtract 4 from stream.offset.
+            if (tagInfo.untilTag) {
```
wayfarer3130: I ended up having to re-read the code - could you rename the tagInfo version of untilTag to isUntilTag so it is clear that this is a boolean indicating it is the is until flag, not the value.  Then I will approve and merge it.

LubosD: Done.

wayfarer3130 (approved):

PR #492, opened 2026-04-23
fix(DicomMessage): parse UN-encoded sequences using Implicit VR LE per PS3.5 §6.2.2

jimOnAir:
## Problem

When a DICOM file encodes a sequence element with `VR=UN` and a **defined length** (valid per DICOM PS3.5 §6.2.2), dcmjs fails to parse it correctly.

### Root cause

DICOM PS3.5 §6.2.2 requires that when a sequence is stored with `VR=UN` and a defined length, its items **must** be parsed using **Implicit VR Little Endian**, regardless of the file's overall transfer syntax.

dcmjs did not implement this rule. It attempted to read the item bytes as Explicit VR, encountered `28 00` as the VR string (the character `(` + NUL), emitted `Invalid vr type ( - using UN`, and stored the value as `[{}]`.

### Consequence

On write-back, the corrupt `[{}]` payload caused a ~843 MB length field to be emitted for the element (e.g. files containing `UDISequence (0018,100A)`), inflating an 788 KB file to 843 MB.

## Fix

In `DicomMessage._readTag()`:

1. Record the **wire VR** (`wireVrType`) immediately after `stream.readVR()`, before any dictionary override changes `vrType`.
2. After reading the element length, if `wireVrType === "UN"` and the payload starts with `FE FF` (the little-endian group bytes of the DICOM item-start tag `0xFFFEE000`), promote the element to `SQ` and switch the recursive read to `IMPLICIT_LITTLE_ENDIAN`.

The `wireVrType` approach correctly handles both:
- **Unknown tags** — where `vr.type` stays `"UN"`
- **Known tags** like `UDISequence (0018,100A)` — where the existing dictionary-override already promotes `vrType` to `"SQ"` before our check runs

## Test

Adds a regression test that constructs a hand-crafted Explicit VR Little Endian buffer encoding `UDISequence (0018,100A)` as `VR=UN` with a defined length, containing one Implicit VR item with `UniqueDeviceIdentifier (0018,1009)`. Verifies:
- The element is promoted to `SQ`
- The inner item is a non-empty array (not `[{}]`)
- The UDI string value is correctly recovered

### DIFF
--- a/src/DicomMessage.js
+++ b/src/DicomMessage.js
@@ -294,7 +294,8 @@ export class DicomMessage {
 
         var length = null,
             vr = null,
-            vrType;
+            vrType,
+            wireVrType = null; // VR as literally encoded in the stream (explicit mode)
 
         if (implicit) {
             length = stream.readUint32();
@@ -318,6 +319,7 @@ export class DicomMessage {
             vr = ValueRepresentation.createByTypeString(vrType);
         } else {
             vrType = stream.readVR();
+            wireVrType = vrType; // remember literal stream VR before any override
 
             if (
                 vrType === "UN" &&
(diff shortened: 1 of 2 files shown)


pieper:
Thanks for the nice PR, thanks.  I was totally unaware of this particular scenario.

I assume this arose in practice with a real datafile?  I'm guessing you can't share it, but maybe you can describe the equipment / vendor that it came from?

jimOnAir:
Hi.
Yes. Main issue was on saving dataset. File size was increased from 700KB to 800MB because of UDISequence was parsed as Array with empty object and then written to file zero padded 843,197,226 bytes value.

Here is data from the image:

```
(0008,0070) | LO | Manufacturer | TERARECON
(0008,1090) | LO | ManufacturerModelName | AQNET
```

jimOnAir:
@wayfarer3130 When you have time, could you please review this PR?

wayfarer3130 (commented):
I have some additional tests on this, plus the fix for the AsyncDicomReader side of things.  Could you give me write access as a maintainer to push to your PR so that I can incorporate the other changes?  Then I can go ahead and approve it after pushing the additional fixes.

wayfarer3130:
I added some additional changes in https://github.com/dcmjs-org/dcmjs/pull/496 to fix it more broadly and have a bit mroe comprehensive tests.

PR #493, opened 2026-04-25
fix: fix charset handling in async reader

LubosD:
As I am switching dicom-curate to use the async parser of dcmjs, I was troubleshooting the fact that processed files suddenly had different checksums. I narrowed it down to this discrepancy:

```
-(0008,0005) CS [ISO_IR 192]                             #  10, 1 SpecificCharacterSet
+(0008,0005) CS [ISO_IR 100]                             #  10, 1 SpecificCharacterSet
```

The dcmjs sync parser always outputs ISO_IR 192 (as it re-encodes strings to UTF-8). The async parser also converts strings to UTF-8, but forgets to override the related tag. This PR resolves the difference.

### DIFF
--- a/src/AsyncDicomReader.js
+++ b/src/AsyncDicomReader.js
@@ -748,6 +748,13 @@ export class AsyncDicomReader {
                     throw Error(`Unsupported character set: ${coding}`);
                 }
             }
+            // Normalize to UTF-8 in the stored value, matching the synchronous
+            // DicomMessage path which always rewrites SpecificCharacterSet to
+            // ISO_IR 192 after decoding (dcmjs always re-encodes output as UTF-8).
+            //
+            // TODO: the original charset value should also be preserved:
+            // it is needed for bulk data decoding.
+            values = ["ISO_IR 192"];
         }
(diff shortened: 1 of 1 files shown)


src/AsyncDicomReader.js
```diff
@@ -748,6 +748,10 @@ export class AsyncDicomReader {
                     throw Error(`Unsupported character set: ${coding}`);
                 }
             }
+            // Normalize to UTF-8 in the stored value, matching the synchronous
+            // DicomMessage path which always rewrites SpecificCharacterSet to
+            // ISO_IR 192 after decoding (dcmjs always re-encodes output as UTF-8).
+            values = ["ISO_IR 192"];
```
wayfarer3130, line 757: We need to keep the original value as well - it is required to have in order to be able to decode bulkdata.   Lets proceed with this for now, but add a comment about it needing to preserve the raw value too.

LubosD: TODO added

wayfarer3130 (approved):

PR #494, opened 2026-05-01
fix(sr): avoid injecting options into Template ContentSequences

### LINKED ISSUE #434: TrackingIdentifier constructor calls super(options) and injects raw options into the ContentSequence

igoroctaviano:
## Summary

Fixes SR template classes that incorrectly called `super(options)` where `Template` extends `Array` via `ContentSequence`. Passing a plain options object into `Array` added a non-DICOM element (the raw `{ … }` object) as the first sequence item, which broke downstream validation and produced invalid content sequences.

Closes https://github.com/dcmjs-org/dcmjs/issues/434  
Addresses review feedback on https://github.com/dcmjs-org/dcmjs/pull/435#issuecomment (extend fix to all affected `Template` subclasses).

## Changes

- **`TrackingIdentifier`**: Call `super()` so the sequence contains only the intended `TEXT` (optional) and `UIDREF` items—no injected options object.
- **Same fix for four other `Template` subclasses** that used `super(options)`:
  - `NormalRangeProperties`
  - `SubjectContextDevice`
  - `MeasurementsDerivedFromMultipleROIMeasurements`
  - `TimePointContext`
- **Tests**: Add `test/TrackingIdentifier.test.js` to lock in correct length and value types for `TrackingIdentifier`.

## Why this is safe

The removed first element was never valid DICOM content; callers that spread these instances (`push(...instance)`) now receive only proper content items. Internal repo usage does not rely on the buggy extra element.

## Checklist

- [x] Build passes (`npm run build`)
- [x] Tests pass (`npm test`)

### DIFF
--- a/src/sr/templates.js
+++ b/src/sr/templates.js
@@ -319,7 +319,7 @@ class MeasurementStatisticalProperties extends Template {
 
 class NormalRangeProperties extends Template {
     constructor(options) {
-        super(options);
+        super();
         if (options.values === undefined) {
             throw new Error(
                 "Option 'values' is required for NormalRangeProperties."
@@ -796,7 +796,7 @@ class SubjectContextSpecimen extends Template {
 
 class SubjectContextDevice extends Template {
     constructor(options) {
-        super(options);
+        super();
         if (options.name === undefined) {
             throw new Error(
                 "Option 'name' is required for SubjectContextDevice."
@@ -1164,7 +1164,7 @@ class VolumetricROIMeasurementsAndQualitativeEvaluations extends _ROIMeasurement
 
 class MeasurementsDerivedFromMultipleROIMeasurements extends Template {
     constructor(options) {
-        super(options);
+        super();
         if (options.derivation === undefined) {
             throw new Error(
                 "Option 'derivation' is required for " +
@@ -1433,7 +1433,7 @@ class MeasurementReport extends Template {
 
 class TimePointContext extends Template {
     constructor(options) {
-        super(options);
+        super();
         if (options.timePoint === undefined) {
             throw new Error(
                 "Option 'timePoint' is required for TimePointContext."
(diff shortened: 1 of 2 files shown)


wayfarer3130 (approved):
Thanks for picking this up.

PR #496, opened 2026-05-13
fix/bill un sq implicit vr ps35622

wayfarer3130:
This is a copy of the code from https://github.com/dcmjs-org/dcmjs/pull/492 with some additional changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "devDependencies": {
         "@babel/core": "^7.17.8",
         "@babel/plugin-transform-runtime": "^7.22.5",
-        "@babel/preset-env": "^7.16.11",
+        "@babel/preset-env": "7.29.5",
         "@babel/register": "^7.13.0",
         "@rollup/plugin-babel": "^6.0.4",
         "@rollup/plugin-commonjs": "^25.0.7",
--- a/src/AsyncDicomReader.js
+++ b/src/AsyncDicomReader.js
@@ -267,6 +267,16 @@ export class AsyncDicomReader {
             // current tag can be cleared.
             stream.consume();
             const tagInfo = this.readTagHeader(options);
+
+            // Stop when the requested tag boundary is reached.  readTagHeader()
+            // has already consumed the 4-byte tag but nothing beyond it, so
+            // stream.offset now points at the first byte after the tag (i.e.
+            // the VR field for explicit-LE).  Callers that need the start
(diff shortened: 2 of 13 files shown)


jimOnAir:
Hello. 
Discussed it with Copilot. Here are his findings:

---

### 1. The byte-peek guard (`FE FF`) is not required by the standard

Both `DicomMessage.js` and `AsyncDicomReader.js` gate the UN→SQ promotion on:

```js
stream.peekUint8(0) === 0xfe && stream.peekUint8(1) === 0xff
```

PS3.5 §6.2.2 does not require a content sniff. The rule is:

> *"…if the Value Representation is UN and Value Length is not FFFFFFFF… the Value shall be read using Implicit VR Little Endian…"*

The peek creates two problems:
- A legitimate `UN` binary blob whose first bytes happen to be `FE FF` will be silently mis-promoted to `SQ`.
- A compliant PS3.5 §6.2.2 sequence that does not start with `FE FF` (e.g. after any kind of stream offset issue, or a future edge case) will be silently skipped.

The conformant check is just:
```js
wireVrType === "UN" && length !== UNDEFINED_LENGTH && length !== 0
```

---

### 2. `length > 0` excludes valid empty sequences

An empty `UN`-encoded sequence (length = 0) is valid per the standard and should be promoted to `SQ` with an empty items array. The current `length > 0` guard leaves it decoded as `UN`, which would corrupt a round-trip (read as `UN []`, written back as `UN`).

---

### 3. Minor: explicit `!implicit` guard asymmetry

`AsyncDicomReader.js` correctly adds `!implicit` as an explicit condition. In `DicomMessage.js` the block is structurally inside the `else` (explicit) branch so it's functionally guarded, but adding `!implicit` explicitly would make the intent defensive and symmetrical across both parsers.

---

### 4. Test gap — unknown tag path

The fixture uses `(0018,100A)` which is in the dictionary as `SQ`, so the dictionary-override path runs first. It would be good to add a test using a private/unknown tag (e.g. `(0009,0010)`) encoded as `UN` with a defined length and a valid item payload, to confirm the fix also works for tags with no dictionary entry.

---

Overall the structure is solid — the `wireVrType` tracking and async propagation are exactly right. The main ask is dropping the content sniff and letting the standard's length-based rule drive the promotion.

igor-nanox:
Any chance of this to be released? It would be extremely helpful. Thanks in advance!

PR #497, opened 2026-05-13
feat: Switch to pnpm for build

wayfarer3130:
This PR uses pnpm for building since there are some issues with yarn that were causing problems.


.npmrc
```diff
@@ -0,0 +1,4 @@
+save-exact=true
+# In GitHub Actions, workflows set PNPM_CONFIG_FROZEN_LOCKFILE=true so `pnpm install`
+# fails when pnpm-lock.yaml is out of date with package.json. Locally, run
+# `pnpm run install:update-lockfile` after dependency changes.
```
sedghi, line 5: it might be good idea to follow the recommendations to mitigate supply chain attacks 

https://pnpm.io/supply-chain-security

pieper (approved):
I have no particular experience with this, so I'm glad to follow your lead on this.  I agree with @sedghi that we should follow any guidance we can on this.

.npmrc
wayfarer3130: Addressed the remaining concerns from that.

PR #502, opened 2026-07-09
feat: add cancellable async DICOM stream reads

jssuttles:
## Summary

This adds a cancellable async byte-stream path for `AsyncDicomReader`.

The main goal is to support metadata-only reads from large DICOM files without forcing callers to read Pixel Data. This is useful for stream/range-based workflows where the caller only needs tags before the image payload.

## Changes

- Add `readFileFromAsyncStream()` to coordinate stream pumping, parsing, and cleanup.
- Add bounded read-ahead so callers can limit how much data is buffered past the parser's current needs.
- Add early-stop support via:
  - `untilTag`
  - `stopOnGreaterTag`
  - `shouldStop`
- Add `stopInfo` so callers can inspect where parsing stopped.
- Support Node streams, browser/Web `ReadableStream`s, and generic async iterables.
- Normalize `untilTag` option values across sync and async readers.
- Preserve source errors even if parsing has already reached an early-stop condition.

## Why this is safe

Existing ArrayBuffer parsing and manual `reader.stream.addBuffer()` usage remain supported.

Stream cancellation is additive:
- Node streams use `destroy()`.
- Web streams use `reader.cancel()` and release reader locks.
- Generic async iterables attempt `iterator.return()`.

`stopOnGreaterTag` follows the DICOM rule that Data Elements are ordered by increasing tag number, and is scoped to the current dataset level so top-level stops do not trigger inside sequence items.

## Notes

Streams can only be stopped between delivered chunks, so callers that need strict no-pixel-byte buffering should choose source chunk/range sizes accordingly.

Deflated transfer syntax behavior is unchanged.

## Checklist

- [x] `pnpm test`
- [x] Node 24 full test suite
- [x] focused async stream tests
- [x] lint and format checks

### DIFF
--- a/docs/AsyncDicomReader-skill.md
+++ b/docs/AsyncDicomReader-skill.md
@@ -116,12 +116,19 @@ const source = fs.createReadStream("image.dcm");
 
 const result = await reader.readFileFromAsyncStream(source, {
     untilTag: TagHex.PixelData,
-    includeUntilTagValue: false
+    includeUntilTagValue: false,
+    streamOptions: {
+        readAheadHighWaterMark: 64 * 1024
+    }
 });
 
 console.log(result.dict);
 console.log(result.stopInfo);
 ```
+
+`readAheadHighWaterMark` applies between source chunks. A single incoming
+chunk is added atomically, so retained bytes may exceed the mark by up to one
+source chunk.
 
 The same API accepts a browser response body:
 
--- a/docs/AsyncDicomReader-skill.md
+++ b/docs/AsyncDicomReader-skill.md
@@ -146,15 +146,17 @@ such as a STOW handler, can reject the read with its own error:
 
 ```javascript
 const readPromise = reader.readFileFromAsyncStream(source, options);
-
-try {
-    await processIncomingRequest();
-} catch (error) {
+const processingPromise = processIncomingRequest().catch(error => {
     reader.pump.abort(error);
-}
+    throw error;
+});
 
-await readPromise; // rejects with the external error
+await Promise.all([readPromise, processingPromise]);
 ```
+
+`pump.abort(error)` rejects a read that is still pending. It cannot change the
+state of a read promise that has already fulfilled, so the processing promise
+must also be awaited as shown above.
 
 Generic async iterables remain supported by the lower-level
 `reader.stream.fromAsyncStream()` API. They are not accepted by
@@ -465,6 +467,20 @@ Reads the entire DICOM file including meta information and dataset.
 
 **Returns:** `Promise<AsyncDicomReader>` - The reader instance
 
+#### `async readFileFromAsyncStream(stream, options)`
+Reads a DICOM file from a cancellable Node stream or browser
+`ReadableStream`. The source is stopped after parsing completes or reaches an
+early-stop condition.
+
+**Options:**
+- All `readFile()` options
+- `streamOptions.readAheadHighWaterMark` (number): Pause source reads between
+  chunks when unread bytes reach this threshold
+- `readerOptions` (Object, static factory only): Options for the reader
+  constructor
+
+**Returns:** `Promise<AsyncDicomReader>` - The reader instance
+
 #### `async readMeta(options)`
 Reads only the file meta information (Group 0x0002).
 
@@ -510,6 +526,10 @@ Reads a single tag header (tag, VR, length).
 - `dict` (Object): Dataset dictionary
 - `stream` (ReadBufferStream): Underlying buffer stream
 - `listener` (DicomMetadataListener): Current listener instance
+- `pump` (Object): Active stream handle exposing `abort(error)`, `failure`,
+  and `finished`
+- `stopInfo` (Object): Early-stop reason, tag/value offsets, available bytes,
+  and loaded end offset
 
 ## Common Patterns
 
@@ -783,7 +803,6 @@ The AsyncDicomReader is marked as preliminary. Future versions may include:
 - Files without DICM preamble
 - Improved streaming for network sources
 - Progress callbacks
-- Cancelable operations
 - More robust error recovery
 
 ## References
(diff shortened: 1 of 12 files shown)


pieper:
This sounds good.  @wayfarer3130 what do you think?

wayfarer3130 (changes requested):
Generally I'm ok with the changes and think they are worthwhile, but there are a few bugs:
1. Non-stream input hangs forever + unhandled rejection (high). In [BufferStream.js:424](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/BufferStream.js#L424), asyncSourceFromStream(stream) is called outside the try/finally, so when it throws (e.g. a caller passes an ArrayBuffer or Buffer — an easy mistake given readFile accepts buffers), setComplete() never runs. readFile() then waits forever in ensureAvailable(132) while the pump's TypeError becomes an unobserved rejection (process-fatal on modern Node). I reproduced this: readFileFromAsyncStream(new ArrayBuffer(16)) times out rather than rejecting. Fix: validate the source synchronously in pumpAsyncStream() and throw from there, before finished is created.

2. untilTag + includeUntilTagValue: true silently never stops the async reader (medium). readTagHeader only sets isUntilTag when !includeUntilTagValue ([AsyncDicomReader.js:777-789](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/AsyncDicomReader.js#L777-L789)), and the read loop has no post-value check equivalent to the sync reader's stop at [DicomMessage.js:114](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/DicomMessage.js#L114). I confirmed with a test: the value is read, but stopInfo stays null and parsing continues to the end of the stream — meaning the source is never cancelled early, defeating the PR's purpose for that option combination. This predates the PR, but the PR advertises untilTag as a stream early-stop mechanism, so it should either work (add a setStopInfo check after the tag is delivered, like shouldStopAfterTag) or throw/document.

Other issues
Hot-path perf regression for all readers/writers: increment() now calls notifyReadAheadListeners() on every single byte-level read/write ([BufferStream.js:362](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/BufferStream.js#L362)), and that method unconditionally allocates a spread copy plus a splice even when there are no listeners — which is always the case for the sync DicomMessage.readFile/write paths. Add if (this.readAheadListeners.length === 0) return; at the top of [notifyReadAheadListeners](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/BufferStream.js#L552-L556).
Per-tag option normalization: readTagHeader ([AsyncDicomReader.js:767](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/AsyncDicomReader.js#L767)) and DicomMessage._readTag ([DicomMessage.js:288](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/DicomMessage.js#L288)) spread-copy options and re-run normalizeTag for every data element, even though _read/readFile entry points already normalized once. On files with thousands of tags this is needless allocation in the hottest loop — normalize at entry points only.
destroyOnAbort: false can hang readFileFromAsyncStream: with that flag, abort() neither destroys nor calls iterator.return() ([BufferStream.js:390-404](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/BufferStream.js#L390-L404)), but shouldAwaitStreamAbort still defaults to awaiting for Node streams ([AsyncDicomReader.js:82-91](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/AsyncDicomReader.js#L82-L91)). If the pump is parked in source.next() on a stalled source (a socket, not a file), finished never settles and the read never resolves despite having succeeded. Factor destroyOnAbort into the await default, or fall back to iterator.return() when not destroying.
stopStreamOnComplete: false orphans the pump: the handle isn't exposed anywhere, so the caller can't abort it later or observe its errors — a later source error becomes an unhandled rejection. Consider storing it (this.pump = pump) or returning it in the result.
isAbortError is identity-only ([BufferStream.js:451](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/BufferStream.js#L451)): correct for fs streams (destroy(err) re-emits the same object), but composed/pipelined Node streams can surface ERR_STREAM_PREMATURE_CLOSE instead, which would make a successful early-stopped read reject. Worth verifying against an HTTP response stream, since range-request reading is the stated use case.
shouldStop fires inside sequence items (it survives the itemOptions spread at [AsyncDicomReader.js:458-464](vscode-webview://1keksvmjsquel6t1hl0jrcbcfp3f39hd7hdtj5ulfchp9nd7pmjm/src/AsyncDicomReader.js#L458-L464)), while untilTag/stopOnGreaterTag are deliberately top-level-only. That asymmetry may be intentional, but it's undocumented and untested — a callback like the test's tagInfo.tag === TagHex.Rows would also trigger on a nested Rows.

As well, the abort mechanism should have a way to tell if this is an abort of the input because it is effectively completed the necessary read, OR it is an abort because of some external processing issue.  The first one should just return the DICOM, while the second one needs to be a promise rejection.  There is a DICOMweb STOW handler already written in static dicomweb that uses just such an abort capability, and it is ending up being hidden here.

jssuttles:
Thanks. Looking into fixing your findings.

jssuttles:
Wasn't sure about your merge/review process - so pinging @wayfarer3130

Issue #505, opened 2026-07-31
unit2CodingValue has no entry for "cm" — centimetre measurements become [arb'U]{cm}


afip94:
### Description

`src/utilities/TID300/unit2CodingValue.js` builds `unitCodeMap` from `knownUnits`. That list contains `cm2` and `cm2/ml`, but **not plain `cm`**.

A length measured in centimetres — the normal case for ultrasound images calibrated through `SequenceOfUltrasoundRegions`, where `PhysicalUnitsXDirection = 3` means cm — therefore falls through to the arbitrary-unit branch:

```js
return {
    CodeValue: `[arb'U]{${units}}`,
    CodingSchemeDesignator: "UCUM",
    CodeMeaning: `arbitrary ${units}`
};
```

The resulting SR carries `CodeValue: "[arb'U]{cm}"`, `CodeMeaning: "arbitrary cm"`. Observed exactly that while writing a Length SR from OHIF for a GE Vivid E95 acquisition (dcmjs 0.52.0).

`cm` is a regular UCUM code, so this looks like an omission rather than a deliberate exclusion.

### Suggested fix

Add it to `knownUnits`, or alongside the explicit `unitCodeMap` entries:

```diff
+unitCodeMap["cm"] = {
+    CodingSchemeDesignator: "UCUM",
+    CodingSchemeVersion: "1.4",
+    CodeValue: "cm",
+    CodeMeaning: "cm"
+};
 unitCodeMap["HU"] = {
```

With that entry present, the measurement round-trips correctly: `10.0429` stored with UCUM `cm`, re-displayed as `10.04 cm`. Without it, `[arb'U]{cm}`.

### Note

A related gap exists in cornerstone3D, where the Length adapter never passes a unit at all, so the `mm` default is used. Reported separately. The two together mean a centimetre ultrasound measurement cannot currently be stored faithfully.

pieper:
That sounds very reasonable.  Do you want to make a PR?

Issue #503, opened 2026-07-09
Sequence-nested string content decoded with Latin-1 instead of SpecificCharacterSet (UTF-8 mojibake)


hromsdahl:
## Summary

String-VR content (TEXT, ST, UT, LO, SH, PN, ...) nested inside a Sequence (SQ) element is decoded using a hard-coded `TextDecoder("latin1")` instead of the decoder derived from the dataset's `SpecificCharacterSet` (0008,0005), even when that attribute correctly declares e.g. `ISO_IR 192` (UTF-8) and appears before the sequence in the dataset. Top-level (non-nested) string attributes in the *same file* decode correctly. The corruption is on the **read** side only — the bytes written for non-ASCII content are correct UTF-8.

Observed on dcmjs 0.50.1 (`dcmjs.data.DicomMessage`/`DicomMetaDictionary`), but the relevant code appears unchanged on `master` as of this writing.

## Minimal reproduction

```js
import dcmjs from 'dcmjs';
const { DicomDict, DicomMetaDictionary, DicomMessage } = dcmjs.data;

function buildBuffer(text) {
  const dataset = {
    SOPClassUID: '1.2.840.10008.5.1.4.1.1.88.33',
    SOPInstanceUID: '1.2.3',
    SpecificCharacterSet: 'ISO_IR 192',
    StudyDescription: text,           // top-level attribute -- NOT nested in any Sequence
    ValueType: 'CONTAINER',
    ContinuityOfContent: 'SEPARATE',
    ContentSequence: [
      { RelationshipType: 'CONTAINS', ValueType: 'TEXT', TextValue: text }, // nested one level in ContentSequence (SQ)
    ],
  };
  const dicomDict = DicomMetaDictionary.denaturalizeDataset(dataset);
  const newDicomDict = new DicomDict({ '00020010': { Value: ['1.2.840.10008.1.2.1'], vr: 'UI' } });
  newDicomDict.dict = dicomDict;
  return newDicomDict.write();
}

const text = 'café';
const buf = buildBuffer(text);
const naturalized = DicomMetaDictionary.naturalizeDataset(DicomMessage.readFile(buf).dict);

console.log(naturalized.StudyDescription);              // "café"      -- correct
console.log(naturalized.ContentSequence[0].TextValue);   // "cafÃ©"     -- mojibake
```

[...]

Issue #484, opened 2026-03-15
Handle SR node encoding without overriding global encoding.


luissantosHCIT:
This is a follow up issue to discussions in #455 in regards to how encodings should be handled. In that PR, I tackled the correction of the value transcoding based on header character set. It was identified that the current character set detection has the potential to corrupt values if multiple character sets are requested (first at the global header and then within an SR node). 

The correct handling would to use "partial specialization subclasses that detect a particular variant and convert the dataset into a form that you would expect according to the standard." Using this pattern, we could temporarily switch the transcoding encoding to handle special cases like SR nodes.

Note: There are many SR objects in the wild switching character sets within nodes.

Issue #479, opened 2026-03-03
Handling raw values with AsyncDicomReader


Iman974:
Hello!

I'm not sure whether this is an unimplemented feature or an intentional design choice, but I noticed that `rawValue` returned by `ValueRepresentation().read()` seems to be discarded in `AsyncDicomReader().readSingle()` and only the formatted value seems to be processed, unlike in `DicomMessage._readTag()` which keeps it. The two places in the code are [here](https://github.com/dcmjs-org/dcmjs/blob/4427728650051fc84e9ace4ae34efa498ec5b07e/src/AsyncDicomReader.js#L718) and [here](https://github.com/dcmjs-org/dcmjs/blob/4427728650051fc84e9ace4ae34efa498ec5b07e/src/AsyncDicomReader.js#L722). This causes the following loss of formatting when parsing with AsyncDicomReader :
<p align="center">
<img width="800" height="500" alt="Image" src="https://github.com/user-attachments/assets/c82b3dff-e258-4ea3-9495-038284619b6b" />
</p>
On this image, the left hand side shows the tags in my original dicom sample file and on the right the file output by something like that :

```javascript
const reader = new AsyncDicomReader();
reader.stream.setData(sampleDicomBuffer); // This function does not exist but pretend it does
await reader.readFile();

// Writing it back (ignoring the incompatiblity with pixel data nesting, say by removing the pixel data entry from the dataset)
const dicomData = new DicomDict(reader.meta);
dicomData.dict = reader.dict;
const outputBuffer = dicomData.write();
// Then download outputBuffer via a Blob
```

I don't have an example to post but I checked with `DicomMessage.readFile()` and then rewriting back - it does keep the proper formatting. If this is not intended, how hard would it be to be able to get the properly formatted (untouched) file back ? I really need to use AsyncDicomReader and not DicomMessage because it exposes the underlying stream properties like the byte offset.

Thank you :)

pieper:
Yes, we do want the behavior to be identical in these cases.  The logic for retaining and raw data should be replicated into the streaming use case.  The idea is that if the javascript version of the value is equal to the re-parsing the raw value then the raw value should be used.

See this discussion: https://github.com/dcmjs-org/dcmjs/issues/398#issuecomment-2269813241

It would be great to have your example in the form of a test to confirm that the streaming and non-streaming paths stay in sync.

PR #512, opened 2026-09-14
feat: dcmjs 1.0 rewrite — event streaming, bounded-memory large files, FHIR, video, DICOMDIR

awatson1978:
## dcmjs 1.0 — a substantial rewrite

This proposes the `awatson1978/dcmjs@development` line as **dcmjs 1.0**,
made possible by the generous support of the National Institutes of
Health and Massachusetts General Hospital. It is large by design — 137
commits, 302 files, roughly **+143,700 / −4,400** — so the intended
entry points for review are the two prose documents, **not** the raw
diff:

- **README → [What changed in 1.0](https://github.com/awatson1978/dcmjs/blob/development/README.md#what-changed-in-10)**
  — the authoritative summary of the new capabilities and the
  API-breaking changes, with a runnable example for each.
- **[BENCHMARKS.md](https://github.com/awatson1978/dcmjs/blob/development/BENCHMARKS.md)**
  — reproducible performance numbers (read / write / roundtrip on the
  committed fixtures, plus warm-vs-cold large-file streaming), each with
  the exact command to regenerate it.

Everything below is a pointer into those two documents.

### What changed (headlines)

- **Event streaming** (`dcmjs.eventStream`) — a source-agnostic,
  SAX-style push parser. A file can be read as a stream of events
  (`startElement`, `value`, `startSequence`, …) flowing source →
  filters → sink, producing the same naturalized metadata without
  materializing the whole dataset. (README → *Event Stream*.)
- **Ultra-large-file support** — pairing the streaming reader with the
  streaming writer keeps peak memory bounded by the largest single
  pixel-data fragment rather than the file size; exercised against a
  21.8 GB video instance while holding ~2 GB. (BENCHMARKS → *Large
  files*.)
- **More correct writing** — element lengths are backpatched during
  write, and the deflated transfer syntax is now actually written
  compressed, fixing a long-standing 0.x bug where such files claimed a
  compression they did not have.
- **DICOMDIR read/write** (`dcmjs.media`), **video encapsulation**
  (Supplement 225 — byte-identical MP4 in and out, no transcoding;
  `dcmjs.encapsulated`), and **instances from decoded pixels**
  (`dcmjs.image`). (README → *Images and Directories*.)
- **FHIR interoperability** (`@dcmjs/fhir`) — bidirectional DICOM ↔ FHIR
  mapping (Patient / ImagingStudy / DocumentReference), also available
  on the event stream. (README → *Fast Healthcare Interoperability*.)
- **Lazy reader deprecated** — the eager reader remains the engine of
  record; the lazy core is behind `DCMJS_CORE=lazy` and scheduled for
  removal.
- **Monorepo** — a pnpm workspace: `dcmjs` at the root, the vendored
  tokenizer in `packages/parser` (private), the FHIR mappers in
  `packages/fhir`, and the docs site in `packages/docs`.

### Breaking changes

- Removed the deprecated `DicomMessage.read` / `readTag` statics and the
  legacy `DICOMWEB` class (use
  [dicomweb-client](https://github.com/dcmjs-org/dicomweb-client)).
- Build tooling is **pnpm-only** (there is no `yarn.lock`); Node.js
  ≥ 22.13. See README → *For Developers*.
- The complete list of API changes is in README → *What changed in 1.0*.

### How to review something this large

The useful review here is at the level of **public API and
capabilities** (the README examples) and **performance claims**
(BENCHMARKS, all reproducible locally). We are happy to split this into
smaller topic PRs, walk through any individual subsystem, or provide
additional context — whatever the maintainers prefer.

### Build & test

```bash
pnpm install && pnpm run build && pnpm test
```

CI targets Node 22 and 24; the parser package carries its own jest suite
and a non-regression gate (`pnpm run bench:parser`).


README.md
```diff
@@ -7,12 +7,64 @@
...
+  pathology slides, surgical video, and large multiframe instances can be
+  gigabytes — larger than available RAM — and the streaming pipeline has
+  been exercised against a 21.8 GB video instance while holding only a
+  couple of gigabytes in memory.
+- **Deprecation of the lazy reader.** dcmjs absorbed the dicom-parser
+  tokenizer and grew a "lazy" reader that records where each element lives
+  and only materializes values when touched. After evaluation, the proven
+  eager reader (read everything up front) remains the default engine of
+  record; the lazy core is deprecated (`DCMJS_CORE=lazy` or
+  `readFile(buffer, { core: "lazy" })`) and scheduled for removal.
```
wayfarer3130, line 40: Please delete this bullet.

The section is "What changed in 1.0". The reader of this section uses dcmjs 0.52.0, the current version on npm. That reader never had a lazy reader:

- `master` has no `src/lazy/` directory, and `master` has no `DCMJS_CORE` string.
- The first commit that adds the lazy core is `bcdfb49`. Four refs contain `bcdfb49`: `development`, `awatson1978/development`, `awatson1978/dcmjsoptimiztaion`, and `origin/dcmjsoptimiztaion`. No commit merges `bcdfb49` into `master`.
- npm shows `latest` as `0.52.0`. npm has no `1.0.0-beta` version.
- `packages/parser` is `"private": true`, so the tokenizer also has no public surface.

The lazy core starts in this branch, and this branch deprecates the lazy core. No release comes between the two events. A deprecation notice tells the reader about a loss, and this reader cannot experience that loss.

Three other places already record the lazy core, and each place is correct:

- `README.md` line 546, the Status section.
- `README.md` line 555, the 1.x backlog.
- `packages/docs/docs/architecture/lazy-core.md`.

The Monorepo bullet at line 61 keeps the one fact that this bullet adds: the repository vendors the dicom-parser tokenizer in `packages/parser`.

wayfarer3130:
I pushed commit `7d4ba70` to this branch. The commit adds one test file. I also found four problems. This comment describes the test file first, and then each problem.

## The new test: a fragment larger than 2 GB

A pixel data fragment item carries a 32-bit unsigned length. A defined-length element carries the same kind of length. A length above 2^31 is therefore legal DICOM, and a producer can write one.

A signed 32-bit read turns such a length into a negative number. A negative length does not throw an error. `fromPart10Stream` computes `fragEnd < fragStart`, the bounds check passes, and the parser emits a short fragment. The parse looks correct to the caller. The new file `test/large-fragment-above-2gb.test.js` guards against that silent corruption of the data.

The default run does not allocate a buffer above 2 GB. Each test file declares a large length, and then the file stops. The whole suite stays at kilobytes, and it runs in the normal CI job.

The tests cover these paths:

- `WriteBufferStream.writeUint32` writes `3000000000` and `0xFFFFFFFE` as unsigned values.
- `StreamingPart10Writer` puts the declared length in the fragment item header without a change.
- `normalizeFragmentBytes` accepts a fragment size above 2 GB.
- `fromPart10Stream` reports `3000000000` (and `4294967294`) in the truncation error. The error text never contains `-1294967296`. The tests check an encapsulated fragment and a native element.
- `fromPart10` refuses the same file.
- Two control tests read a small fragment from the same builder. A failure above is therefore about the length, and not about a malformed file.

I also made a mutation test. I changed `getU32` in `src/eventStream/fromPart10Stream.js` from `getUint32` to `getInt32`. Four of the new tests then failed. I restored the source file, and `git diff src/` is clean.

One test allocates 3 GB. Jest skips that test, unless you set `DCMJS_TEST_HUGE_FRAGMENT=1`. I ran it with that variable, and it passed.

## Problem 1: `DicomMessage.readFile` hides a truncated fragment

Give `DicomMessage.readFile` a 250-byte file whose one fragment declares 3000000000 bytes. The function does not throw an error. The function allocates 3 GB, it copies the 16 bytes that the file supplies, and it fills the rest with zeros. The caller gets a buffer that looks complete.

`fromPart10` and `fromPart10Stream` both refuse the same bytes. The length arithmetic in `DicomMessage.readFile` is correct, because the buffer is `3000000000` bytes and not a negative or clamped value. Only the truncation goes unreported.

The skipped test records this behaviour. The comment in that test says clearly that the test does not endorse the zero padding.

## Problem 2: four test files read a fixture with the wrong offset

Seven tests fail on my machine. They failed before my commit, and they fail when I run each file alone. Node v24.20.0 on Windows, `pnpm test`.

The cause is the Node Buffer footgun, and issue #311 already describes it. Inside Jest, `fs.readFileSync("test/sample-sr.dcm")` returns a Buffer with `byteOffset` 17232 in a 65536-byte pool. The tests then read `.buffer`, which starts at the pool, and not at the file. The DICM prefix is not at offset 128 of that pool, so the reader throws `Invalid DICOM file, expected header is missing`.

The `byteOffset` depends on the state of the pool, so the failure is not deterministic. These three sites need the fix:

- `test/anonymizer.test.js:44` — `fs.readFileSync("test/sample-sr.dcm").buffer`
- `test/data.test.js:1007` — `file.buffer`
- `test/eventStream/StreamingPart10Writer.test.js:11-12` — `.buffer.slice(0)`

Each site must honour the offset:

```js
const b = fs.readFileSync(file);
const arrayBuffer = b.buffer.slice(b.byteOffset, b.byteOffset + b.byteLength);
```

I confirmed the cause. With `.buffer` the read throws; with the sliced buffer above the same file reads correctly.

## Problem 3: a Windows path separator empties a test table

`test/eventStream/streamEquivalence.test.js:453` fails with `.each` called with an empty Array of table data.

`ALL_FIXTURES` builds each key with `path.relative(REPO_ROOT, fullPath)`. On Windows that key uses a backslash. The `SAMPLE_1BYTE` list at line 377 uses a forward slash. The filter therefore matches nothing, and Jest rejects the empty table. A `split(path.sep).join("/")` on the key solves this problem.

## Problem 4: the statement about the 2 GiB limit is version dependent

`BENCHMARKS.md:121` says that Node cannot put more than 2 GiB in one buffer, and line 148 repeats the statement.

On Node v24.20.0 a 2 GiB+ `ArrayBuffer` allocates correctly, and `buffer.constants.MAX_LENGTH` is `9007199254740991`. The limit depends on the version of Node. Please add the version of Node that you measured, so that a reader does not apply the statement to a newer version.

## A note about CI

The workflow runs on `awatson1978/dcmjs-1.0` do not start. Each job reports `The job was not started because your account is locked due to a billing issue.` The run for my commit is `35004028128`. There is therefore no CI result for this branch, and I ran the tests only on my machine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pieper:
I asked Claude for a review and got the following.  I'm not sure it's all valid but I'll put it here for reference.

< start of AI >

Thank you @awatson1978, and @sedghi for the work this branch started from. There is a lot of real value here. Bounded-memory streaming is something we have wanted since #444. I pulled the branch, built it, ran the tests, and read through it. Here is where I've landed.

## What's valuable

- **Backpatched element lengths in `Tag.write`.** Removing the per-element temporary stream plus concat looks like the source of the 2–4× write speedup.
- **Deflated output that is actually compressed.** This fixes a long-standing bug where the transfer syntax claimed compression that wasn't there.
- **The ISO 2022 charset decoder (`src/charset`)** and passing the charset down into sequence items (#503).
- **Many small, well-targeted fixes:**
  - the `readUint16Array` off-by-one
  - `SplitDataView` ignoring `byteOffset` (#311)
  - DS commas (#287)
  - NUL padding (#130)
  - meta group length recovery (#338)
  - RLE fragments (#340)
  - warning only once for unknown VRs (#457)
  - and others
- **The regression suites harvested from old issues.**
- **The event stream idea itself,** which is worth a proper design discussion.

## Why I can't review or merge it as one PR

**Size and makeup.** It is +143k lines, but by my count roughly:
- 53% is generated schema artifacts (`schema/naturalized.schema.json`, `src/schema/naturalizedRules.js`, `types/dcmjs-schema.d.ts`)
- 18% is tests and 10% is lockfile, mostly the Docusaurus tree
- 8% is docs and process files (`CLAUDE_*.md`, `.superpowers/`, HTML reports)
- about 6.5% is new hand-written source

The history also includes a rewrite and about 29 MB of blobs, so I would not merge it with a merge commit or change the default branch.

**Review burden.** Using AI to move dcmjs forward is welcome; I plan to do that myself. But 108 of the 117 non-merge commits are AI co-authored, and a maintainer can't meaningfully review a change this size in one pass. We need it in pieces that a human can reason about.

**Architecture: more read engines, not fewer.** Master has two readers (`DicomMessage._read` and `AsyncDicomReader`). This branch has five:
- the eager reader
- `AsyncDicomReader`
- the deprecated lazy core
- `decodeCore`/`fromPart10` over the vendored parser
- `fromPart10Stream`

Some divergences between them are already documented in the tests. The lazy core was made the default and later deprecated ("stakeholder pivot"), but it is still in the branch, along with the passthrough and zero-copy window complexity it added to `SplitDataView`/`BufferStream`. Since no release ever shipped it, it should just be removed.

**Overlap with open work.** The branch doesn't reconcile with work that is already in flight:
- #423 (the 1.0 plan: types, removing proxies, ESM)
- #502 (cancellable async reads)
- #481 (async deflate)
- #492/#496 (UN-encoded sequences; this branch covers only the undefined-length case)
- #510 (anonymizer keywords, a line-for-line conflict)
- #511 (OL/OV/SV)
- #455/#483 (VR/encoding refactor)

It also adds a parallel `EventStreamListener` instead of extending `DicomMetadataListener`.

**Scope.**
- **FHIR mapping** isn't DICOM.
- **DICOMDIR** is listed as a non-goal in the README's Goals section.
- **The docs site, video/MP4, PDF and image builders, and the schema catalog** would all be better as separate, optional packages.

As it stands, all of these hang off the default export, so bundlers can't drop them: in my build `dcmjs.min.js` grew about 19% and `dcmjs.es.js` about 36% compared with 0.52.0.

## Behavior changes that should be called out as breaking

Some of these are improvements, but downstream users (OHIF, `@cornerstonejs/adapters`, dcmjs-dimse, dicom-curate) need to know:

- **Private tags now survive naturalize → denaturalize.** This is good in general, but it has privacy consequences for any de-identification pipeline that relied on them being dropped.
- **`xs` VRs are resolved from PixelRepresentation.** Please double-check the LUT Descriptors (0028,1101–1103) and (0028,3002); PS3.3 says the first and third values are always unsigned.
- **DS strings that don't match the PS3.5 grammar now become `null`.** Invalid characters used to be stripped.
- **The anonymizer's default `tagNamesToEmpty` list changes** substantially.
- **Other changes:** the logger now uses a child logger and patches `getLogger`; `uid()` generation changed; the `SplitDataView` listener signature changed; `isPrivateCreator` is narrower.

Also, "pnpm-only, Node ≥ 22.13" is already on master (#497), so it isn't a 1.0 change.

## Build and test notes

- **Local run (Node 22.20):** 1,713 passed, 1 failed, 11 skipped, in about 32 minutes. Master in the same environment runs 316 tests in about 22 seconds. The failure (`async-data.test.js › private tags are read correctly`) passes on its own, so it looks like a flake under load.
- **`streamEquivalence.test.js` alone takes about 21 minutes in CI.** That matrix should be a nightly job, not a PR gate.
- **PR CI:** `lint-and-format` fails (8 ESLint errors), `audit` fails (likely the Docusaurus dependencies), and `test (24)` fails. The Node 24 `fs.readFileSync(...).buffer` failures also affect master, so I'll fix that separately.
- **Two existing tests** (`test_multiframe_1`, `test_oneslice_seg`) moved behind a non-blocking network-fixture job, which weakens the main gate.
- **`packages/parser` licensing:** it vendors dicom-parser without its MIT LICENSE or copyright notice. Committed DICOM fixtures also need provenance and license information.
- **`BENCHMARKS.md`:** it was run on Node 25, which isn't a supported LTS. The large-file comparison is against 0.52's whole-file `readFile` instead of its `AsyncDicomReader`. The `video48-*` fixtures aren't available to reproduce the results.

## Proposed path forward

I'd like to treat this branch as a source of patches and a design proposal. In rough order:

1. **Small bug fixes,** each as its own PR with the matching `test/issues` regression test.
2. **Anonymizer keyword fix,** coordinated with #510 and landed with a changelog note. **UN-sequence handling,** combined with #496.
3. **Deflate-on-write,** with the dclunie deflate fixtures (with license) and coordinated with #481.
4. **The backpatch writer,** with the byte-exact writer tests.
5. **The ISO 2022 charset module,** coordinated with #455.
6. **Separate discussions** (not cherry-picks) for the `xs` resolution, private-tag round-tripping, and the logger change.
7. **The event stream as an RFC** before code, covering how it relates to `AsyncDicomReader`, `DicomMetadataListener`, #482 and #502. Then the streaming reader and writer as follow-up PRs.

**Separate repositories or packages:** `@dcmjs/fhir`, the docs site, DICOMDIR/video/PDF/image builders, and the schema catalog.

**Drop:** the lazy core with passthrough and zero-copy windows, the AI process files and HTML reports, and the Docusaurus template assets.

A few requests:
- Each split PR should have green CI (lint, audit, Node 22 and 24) and a test time close to master's.
- Please add grant or contract numbers to the NIH/MGH acknowledgement.
- @sedghi, since this branch builds on your `dcmjsoptimiztaion` work, I'd like to hear whether the direction it took matches your intent.
- Please check the behavior changes above against `@cornerstonejs/adapters`, OHIF and dcmjs-dimse.
- Please rerun the benchmarks on a supported Node LTS, against `AsyncDicomReader`, with fixtures others can download.

## Longer term

I'm going to write up a broader 1.0 direction in #423. The rough outline:
- A hard API break, in TypeScript.
- Dictionaries and IOD tables generated from the DICOM standard's DocBook source, rather than from our own dictionary.
- Smaller standalone repositories that compose, rather than one large package.

A lot of what's in this branch could land in that structure. I'd welcome your participation there, along with @wayfarer3130, @jssuttles, @luissantosHCIT, @jimOnAir, @cpruijsen and everyone else with work in flight, so we converge on one plan.

Thanks again for the effort. There is a lot here we can use.

< / end of AI >

wayfarer3130:
Found a number of issues in the overall PR - lots found by clod, and I'm still going through changes that group the functionality, but here is the current set:
[PR512-REVIEW-MAP.md](https://github.com/user-attachments/files/32256059/PR512-REVIEW-MAP.md)

sedghi:
@pieper i will review the pr when it is ready, I see it is in draft and ci is failing so i guess it is not yet?

awatson1978:
@pieper  Totally understood on not wanting this all in a single PR.  This initial PR was intended more to document the surface area and scope of proposed changes, and then figure out how we want to stage it.  Just as an FYI, there are actually 44 pull requests in my working dcmjs repo copy; but it didn't get forked from your original correctly; hence the squashing them into a new repo.  
https://github.com/awatson1978/dcmjs/pulls

Regarding the many small, well target fixes - by my count, this PR addresses the following 27 open issues on your repo.  Once we had the event streaming in place, all these extra items were pretty straight forward to address.  

| Issue(s) | Symptom | Fix (pinned regression test) |
| --- | --- | --- |
| #345 | 103/221 anonymizer names dead — PHI survived `cleanTags` | all 103 → real dictionary keywords; scrubs ~103 more attributes |
| #287 | comma-decimal DS parsed 1000× wrong | DS validated vs PS3.5 grammar; invalid → null + warning, `_rawValue` kept |
| #503 | eager reader mojibaked SQ-nested strings (latin1 hard-coded) | item substreams inherit the dataset decoder; all paths agree |
| #311 | pooled Node `Buffer` view parsed the **wrong file** | `SplitDataView.addBuffer` honors `byteOffset`/`byteLength` |
| #340 | RLE frames re-fragmented on write (corrupted cine) | `fragmentMultiframe` never splits frames under RLE (PS3.5 Annex G) |
| #338 | wrong `(0002,0000)` → internal errors / silent garbage | meta group length validated/corrected by structural walk |
| #130 | NUL padding survived into string values | read-side trims strip trailing `0x00`; `_rawValue` keeps originals |
| #388 / #215 | private tags dropped on denaturalize | unmapped elements round-trip via `_vrMap` (recursive in SQ); symmetric `registerTag()` |
| #373 / #284 / #454 / #484 | ISO 2022 multi-charset unsupported / mojibake | new `src/charset/iso2022.js`; all read paths share it; dclunie corpus = pydicom refs |
| #93 | meta-less DIMSE datasets unreadable | `readFile(buffer, { allowMissingHeader: true })` (syntax auto-detected) |
| #42 | `[null]` denaturalize TypeError | null entries pass through |
| #356 | reserved elements 0x01–0x0F misclassified as private creators | `isPrivateCreator` tightened to PS3.5 7.8.1 (0x0010–0x00FF) |
| #417 | context-free "Not a number: undefined" on write | write errors annotate tag + VR (`error.cause` chained) |
| #61 | 2.25 UIDs were random digits, not UUID-derived | `uid()` derives from RFC 4122 v4 UUID (PS3.5 B.2) |
| #479 | `AsyncDicomReader` discarded `rawValue` | `readSingle` threads `_rawValue` like the eager reader |
| #398 | JSON model truncated DS/IS ints > 2^53 | PS3.18 F.2.3.1 number-or-string (number when lossless) |
| #204 | eager/stream PixelData shape divergence | streaming listeners merge BOT-window fragments per frame |
| #368 | `xs` VR ignored PixelRepresentation | US/SS resolved via (0028,0103) in all read paths |
| #297 | import-time latin1 `TextDecoder` dependency | guarded construction + pure-JS latin1 fallback |
| #457 | per-lookup VR-fallback warn spam | warn-once-per-type |
| #477 | docs described a nonexistent `setData()` API | corrected to real `addBuffer`/`setComplete` |
| #505 | `cm` mis-coded as an arbitrary unit in SR | `cm`/`m`/`um` added to the UCUM map |
| #363 | "Invalid tag in sequence" on Philips MR files | UN + undefined length → implicit-VR SQ (PS3.5 §6.2.2) |

In the near term, what I'm gathering as immediate tasks:

- [ ] Remove the lazy core (not just deprecate it).
- [ ] Remove the AI process files and the HTML report.
- [ ] Add dicom-parser's MIT license, and a provenance/license note for the committed fixtures.
- [ ] Take the event stream to an RFC in #423 before any reader/writer PRs.
- [ ] Get every split PR green on lint, audit, and Node 22 and 24, with a runtime close to master's (the long `streamEquivalence` matrix becomes a nightly job).
- [ ] Rerun the benchmarks on Node 22 and 24, add `AsyncDicomReader` as the comparison, and publish downloadable fixtures.
- [ ] Check the flagged behavior changes against `@cornerstonejs/adapters`, OHIF and dcmjs-dimse, and document each as breaking.
- [ ] Double-check the `xs` LUT-descriptor case you raised (0028,1101–1103 and 0028,3002).

On scope, I mostly agree: FHIR, DICOMDIR, video, image, PDF, the docs site, and the schema catalog should be **separate, optional packages**. That's the right fix for the bundle growth you measured — they were riding the default export. None of them belong in the core reader.

The one place I feel compelled to stand my ground on is on treating FHIR and DICOMDIR as outside the mission. I would argue these are the patient's side of imaging:

- **FHIR — not DICOM — is what the government mandated** for EHI export and patient access, and MACRA points at FHIR for billing. An imaging library that can't bridge DICOM to FHIR can't serve either of those use cases.
- **DICOMDIR is how patients actually receive their images** — it's the index on the CD/DVD/USB they're handed. Reading it is patient access.
- there's a growing list of DICOM updates that are referencing FHIR nowadays.  [CP-2034](https://dicom.nema.org/medical/dicom/CP/cp2034_02_FHIRStorageAndRetrievalSequences.pdf?utm_source=chatgpt.com), [PS3.17](https://dicom.nema.org/medical/dicom/current/output/pdf/part17_changes.pdf?utm_source=chatgpt.com), [CP-2196](https://dicom.nema.org/medical/dicom/CP/cp2196_01_AddFHIRCodingSchemeURLType.pdf?utm_source=chatgpt.com), and others.  

So I'd like to:

1. Keep FHIR and DICOMDIR as **first-class, supported packages** in the dcmjs-org ecosystem, not orphaned.
2. **Strengthen the README Goals** to name patient access explicitly — EHI export, access to historical images, billing support — alongside the clinician use cases.
3. Hold them to the same bar as everything else (green CI, tests, changelog).

.superpowers/sdd/task-J2-report.md
```diff
@@ -0,0 +1,166 @@
+# Stage J2 Report — Repoint `readFileLazy` onto decodeCore
```
sedghi, line 1: please remove these

docs/AsyncDicomReader-skill.md
```diff
@@ -72,29 +74,30 @@ async function readDicomFileWithOptions(arrayBuffer) {
...
-  addTag(tag, tagInfo) {
-    console.log(`Processing tag: ${tag} (${tagInfo.name})`);
-    super.addTag(tag, tagInfo);
-  }
-  
-  value(val) {
-    // Process each value as it's read
-    super.value(val);
-  }
+    addTag(tag, tagInfo) {
```
sedghi, line 80: this has empty spaces added or removed some places, please minimize changes

.superpowers/sdd/task-J5-report.md
```diff
@@ -0,0 +1,88 @@
+# Stage J5 Report — Phase 1 (slice J) documentation
```
sedghi, line 1: we should remove all the .superpowers

docs/AsyncDicomReader-skill.md
```diff
@@ -7,63 +7,65 @@ large DICOM files efficiently by reading and processing data incrementally rathe
...
 ## Key Features
 
-- **Asynchronous/Streaming**: Reads DICOM files incrementally using async/await patterns
-- **Memory Efficient**: Uses buffer streaming with automatic clearing to reduce memory footprint
-- **Multiple Transfer Syntaxes**: Supports Explicit Little Endian, Explicit Big Endian, and Implicit Little Endian as well as compressed syntaxes.
-- **Pixel Data Handling**: Handles both compressed and uncompressed pixel data
-- **Sequence Support**: Properly parses DICOM sequences with defined and undefined lengths
-- **Character Set Support**: Automatically handles different character encodings, but not multiple character encodings.
-- **Error Handling**: Configurable error handling for malformed files
+-   **Asynchronous/Streaming**: Reads DICOM files incrementally using async/await patterns
```
sedghi, line 10: a skill should be under .skills by the standard, ask Mo he knows the standard and can guide you

docs/test-report.html
```diff
@@ -0,0 +1,82 @@
+<!doctype html><html lang="en"><head><meta charset="utf-8"><meta name="viewport" content="width=device-width,initial-scale=1">
```
sedghi, line 1: these files seem to be temporary measures and should not be part. ofthe main pr

generate/buildCatalog.mjs
```diff
@@ -0,0 +1,219 @@
+// D22 catalog builder — pure (no I/O, no import.meta), so the CI gates can
```
sedghi, line 1: I can immeditely recognize the opus 5 weird language here, please ask Fable or someone else (not opus) to revisit the language for comments and follow plain english

packages/fhir/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "@dcmjs/fhir",
+    "version": "0.0.0",
+    "private": true,
+    "description": "DICOM → FHIR R4B mapper (Patient + ImagingStudy from naturalized Part 10 datasets). The FHIR sink for the dcmjs pipeline.",
```
sedghi, line 5: What is R4B?

packages/fhir/src/index.js
```diff
@@ -0,0 +1,114 @@
+// packages/fhir/src/index.js
+//
+// @dcmjs/fhir — FHIR both ways.
```
sedghi, line 3: I kind of agree with Steve that FHIR is out of scope for this PR which is dcmjs upgrade to 1.0, we can decide later if @dcmjs/fhir belongs here or not

packages/parser/bench/bundle-gate.mjs
```diff
@@ -0,0 +1,263 @@
+/**
+ * Bundle / side-effect gate for @dcmjs/parser (REWIRING-PLAN.md, R8 step 2).
```
sedghi, line 2: the final merged code should not reference the plan in its docs

packages/parser/src/parseDicom.js
```diff
@@ -0,0 +1,177 @@
...
+import DataSet from './dataSet.js';
+import littleEndianByteArrayParser from './littleEndianByteArrayParser.js';
+import readPart10Header from './readPart10Header.js';
+import sharedCopy from './sharedCopy.js';
+import * as byteArrayParser from './byteArrayParser.js';
+import * as parseDicomDataSet from './parseDicomDataSet.js';
+
+// LEE (Little Endian Explicit) is the transfer syntax used in dimse operations when there is a split
+// between the header and data.
+const LEE = '1.2.840.10008.1.2.1';
```
sedghi, line 13: these probably exist somewhere in dcmjs as a central file that can be imported here

packages/parser/testImages/deflate/README.md
```diff
@@ -0,0 +1 @@
+http://www.dclunie.com/images/compressed/deflate_tests_release.tar.gz
```
sedghi, line 1: please make sure all the .dcm files getting added have a data reference like this  https://github.com/OHIF/Viewers/blob/master/DATACITATION.md

awatson1978:
Thanks, Bill (@wayfarer3130) — copy all of this.  Happy to keep your >2 GB-fragment test (`7d4ba70`) in place as an anchor, and start through your other items as a triage list.  Here's how I'm planning to work through the list. Detailed, per-finding replies will follow as I verify each one, but at a high level:

The confirmed bugs I'll just fix, each in its own small PR with the regression test sitting right next to it. I'll take the HIGH, data-loss ones first — the >2 GB truncation in `readFile`, the `BufferStream` byteOffset (#1), the UV-null write (#3), the private-tag fallback (#4), the little-endian header writer (#5), the deflate deadlock (#6), the dropped source error (#7), the FHIR DocumentReference loss (#8), the JPEG overread (#9), and the missing `NumberOfFrames` (#10). Two of them (#20 and #21) live inside the lazy core, so they'll just disappear once I pull that out.

The two you flagged as improvements, I'd like to keep — the tighter private-creator range (#26) and the wider anonymizer list (#2). Rather than bury a leniency switch inside the parser, I'll write each up in the changelog and offer the old behavior as an opt-in filter you can compose in. On your Z-vs-X question for the six sequences: right now `cleanTags` zeroes them out (that's Z), and the table calls for removing them (X). I'll check it against Table E.1-1 and document whichever way we go.

On #24 (the meta group length): you're right that it's leniency, not strictness. I'll make that explicit, and I'll close the gap you found where an overstated length quietly pulls dataset elements into `meta` with no warning at all.

And your four named problems plus the README bullet — all accepted: the Node-24 `.buffer` offset fix at the three sites, the Windows `path.sep` empty-table, the zero-pad-on-truncation behavior, the version-specific 2 GiB note in BENCHMARKS, and dropping the "lazy reader deprecation" bullet (you're right, no released version ever had it).

If I dig into one of these and come to think it's actually wrong, I'll come back with a reproduction rather than a reflexive "fixed." 

All that being said, before I start, I'd like a game plan in whether we should start this before or after the PR staging.  It's sounding like we put the event-stream in first, then apply these fixes you just found; then the PRs for the supporting packages (FHIR, DICOMDIR, video, image, pdf, schema catalog, validator, etc).

sedghi (changes requested):
I did a quick glance and had some comments. The blocker for review is the english style anthorpic models use to comment (more recently) which just gives headache to read. We need to revisit the language.

On the structure. I"m wondering if it makes sense to have @dcmjs/fixtures to have all dicom files and test images etc etc and @dcmjs/tests (not published) that we have tests. Right now fixtures and tests are all spread in many folders

Great first draft, great work

packages/fhir/package.json
awatson1978: FHIR R4B is the latest version of FHIR supported by Medicare/Medicaid for patient data access, EHI export, billing, etc.

.superpowers/sdd/task-J5-report.md
awatson1978: Already working on it.

Issue #311, opened 2022-09-17
Trouble doing basic loading


vjau:
Hi, 
i'm trying this library for the first time and have trouble with basic operations, trying to run this in node.

With this code and provided test dicom 
```js
const dcmjs = require("dcmjs");
const fs = require('fs');
fs.readFile('IM-0001-0001.dcm', (err, data)=>{
  let dicomData = dcmjs.data.DicomMessage.readFile(data);
});
```
I get following error
```js
this.view = new DataView(this.buffer);
                        ^

TypeError: First argument to DataView constructor must be an ArrayBuffer
```
I am not an expert in node binary handling, but i suppose a node `Buffer` should be accepted as an `ArrayBuffer`.

Or is this project only for the browser ?

Thank you.

pieper:
The code should work fine in node or browser.  Try with one of the test datasets and if you still have issues post the full stack trace.

vjau:
The dicom file used is the one provided in your ```examples/data``` folder.

Stacktrace is here
```
/home/vincent/dev/temp/dcmjs-try/node_modules/dcmjs/build/dcmjs.js:829
            this.view = new DataView(this.buffer);
                        ^

TypeError: First argument to DataView constructor must be an ArrayBuffer
    at new DataView (<anonymous>)
    at new BufferStream (/home/vincent/dev/temp/dcmjs-try/node_modules/dcmjs/build/dcmjs.js:829:18)
    at ReadBufferStream._createSuperInternal (/home/vincent/dev/temp/dcmjs-try/node_modules/dcmjs/build/dcmjs.js:632:25)
    at new ReadBufferStream (/home/vincent/dev/temp/dcmjs-try/node_modules/dcmjs/build/dcmjs.js:1183:21)
    at Function.readFile (/home/vincent/dev/temp/dcmjs-try/node_modules/dcmjs/build/dcmjs.js:3579:21)
    at /home/vincent/dev/temp/dcmjs-try/main.js:4:43
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:71:3)
```

I have made a repro here : https://github.com/vjau/dcmjs-try

Node v14.20.0
Ubuntu 20.04

pieper:
Thanks for making the sample.  Did you try changing this line to `data.buffer`?

https://github.com/vjau/dcmjs-try/blob/1283caae8a16a37bfe5a4cc513297eaa0c3f7e33/main.js#L4

[...]

PR #518, opened 2026-09-16
docs: release plan for landing 1.0 as staged package PRs, plus a pipeline smoke test

awatson1978:
This is the first small PR of the staged approach we discussed for #512. It has two jobs: put the release plan in the repo where everyone can mark it up, and prove that CI actually runs.

**RELEASE_PLAN.md** lays out the whole approach: which packages we're splitting out (`@dcmjs/core`, `@dcmjs/parser`, `@dcmjs/schemas`, `@dcmjs/tests`, `@dcmjs/fixtures`, and so on), how the per-package staging branches work, the test-first sequence for core, what happens to each of the 33 findings from Bill's review map, and how publishing works. It's meant to be read top to bottom, and section 1 sums up the whole idea in a couple of paragraphs.

**The smoke test** is deliberately trivial: two assertions on the `sample-sr.dcm` file that's already in the repo. Its only purpose was to make the checks run.

Running them turned out to be worth it, because the pipeline caught two real problems on its first try:

1. **Node 24 fails 7 existing tests, on master too.** When Node reads a small file it sometimes hands you the file inside a larger shared memory buffer. Grab `.buffer` off that and you get the whole shared buffer, with your file sitting somewhere in the middle — so the parser goes looking for the DICM marker in the wrong place. Node 24 pools more aggressively than 22, which is why the same tests pass on one and fail on the other. Bill and Steve both flagged this in the #512 review, and it's the same family as #311. I added a `readFileAsArrayBuffer()` helper to `testUtils.js` and pointed all the file-reading test code at it.

2. **A flaky temp-directory race.** Two jest workers both check whether the shared temp directory exists, both find it doesn't, and the slower `mkdir` dies with EEXIST. `mkdirSync` with `recursive: true` doesn't care whether the directory is already there, so the check-then-create is gone.

Merging this publishes nothing and touches no library code. It gets the plan into the repo where follow-up PRs can cite it, and leaves the `1.0-beta` gate green, with all eight checks passing, on Node 22 and 24, for both the push and PR runs.


sedghi:
I think the first pr is to fix the ci. to publish beta channel @wayfarer3130

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| `@dcmjs/parser`                | The low-level tokenizer core (adapted from [dicom-parser], with its MIT license carried along)                                                                                                                                                         | `packages/parser/`                                                                                                                                                                                                                                                 | No — bundled inside `@dcmjs/core`'s build (recommendation; final call before the core branch merges) |
```
wayfarer3130: The dicom-parser should be dropped entirely from the 1.0 stream - that was something you were evaluating as the lazy parser and isn't to be carried forward.

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| `@dcmjs/core`                  | Reads and writes DICOM files, including the new event-stream reader that handles files larger than memory                                                                                                                                              | `src/eventStream/`, `src/core/`, the reader/writer classes (`DicomMessage`, `DicomDict`, `ValueRepresentation`, `BufferStream`, `SplitDataView`, `Tag`, `DicomMetaDictionary`), `src/charset/`, `src/constants/`, the runtime dictionary files, `AsyncDicomReader` | Yes                                                                                                  |
```
wayfarer3130: Suggest this contain the ValueRepresentation, split data view, tag, charset, but neither readers nor writers, nor the async dicom reader
Lets put the legacy stuff in @dcmjs/legacy
and the new reader stuff in
@dcmjs/parser (NOT the parser you have below)
@dcmjs/generator for the output

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| `@dcmjs/dicomdir`              | Reads and writes DICOMDIR, the index file on the CDs and USB drives patients receive                                                                                                                                                                   | `src/media/`                                                                                                                                                                                                                                                       | Yes                                                                                                  |
```
wayfarer3130: Lets call this media if it is media, otherwise call it dicomdir

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| `@dcmjs/pdfs`                  | Wraps PDF reports into DICOM instances and unwraps them again                                                                                                                                                                                          | `src/encapsulated/encapsulatedPdf.js`                                                                                                                                                                                                                              | Yes                                                                                                  |
```
wayfarer3130: I think we should have a package that has data helpers for the bulk data generally - that would be a single package that has pdf, video and other data helpers.
Adding too many sub-packages needs to be balanced with having too few and I think this breakdown is a bit too granular.

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| `dcmjs` (the existing package) | Becomes a thin wrapper: the familiar 0.x surface (structured reports, derivations, adapters, utilities, anonymizer) plus a re-export of `@dcmjs/core`. It does **not** bundle the optional packages, which returns the bundle to roughly its 0.52 size | root `src/` minus everything above                                                                                                                                                                                                                                 | Yes                                                                                                  |
```
wayfarer3130: Build a dcmjs package, but lets not publish is quite yet - or perhaps publish it for now as @dcmjs/dcmjs so that both packages can be imported just by changing the import label.  When ready, we will start publishing actual dcmjs builds.

wayfarer3130 (changes requested):
Can you update package names/details as noted in the PR.

The FIRST publish doesn't have to be in the first commit - we can move things around until we are ready to try pushing.

RELEASE_PLAN.md
awatson1978: Yes, I very much agree.  Bumping that up next.

RELEASE_PLAN.md
awatson1978: Ah, we can have packages that bundle other packages.  I was thinking a `@dcmjs/media` package.

RELEASE_PLAN.md
```diff
@@ -0,0 +1,307 @@
...
+| 4    | Core test bank onto `release/1.0-beta-core`                                                            | The tests that define what core must do land **before** the code that does it — see section 7.                                                                                                          |
+| 5    | The 33 review findings, fixed one PR each, then `@dcmjs/core` itself                                   | The heart of the release.                                                                                                                                                                               |
+| 6    | `@dcmjs/dicomdir`, `@dcmjs/pdfs`, `@dcmjs/video`, `@dcmjs/fhir`                                        | Independent of each other once core exists; can proceed in parallel.                                                                                                                                    |
```
sedghi: I think Steve wants fhir out for now until later, please double check with him @pieper

RELEASE_PLAN.md
awatson1978: Confirmed.

RELEASE_PLAN.md
```diff
@@ -0,0 +1,352 @@
...
+| `@dcmjs/mcp`                        | An MCP (Model Context Protocol) server that lets AI assistants work with DICOM data through dcmjs: reading metadata, anonymizing, converting between representations                                                                                                                                                                                         | stretch goal                                                                                                                                                                                                                             | TBD                                                                                                                                                                                                                                              |
```
sedghi, line 86: mcp requires a server, how does it fit into dcmjs scope? can we not add anything new and cut scope even for PRs and features

RELEASE_PLAN.md
wayfarer3130: It won't be part of the core dcmjs, just as one of the extra libraries - that keeps it available/close, but separates it as desired.

